Issue #734, opened 2023-03-30
Fixes 5 | Team, Task, Settings pages


KostyaEver:
- [x] In Settings - there is NOT 'back' arrow and The dashboard in the page path does not work (the user should be able to go back)
- [x] Team page - we don't have the ‘Status’ dropdown on the user car under the title we need to add it ONLY for the Team page (on the User Task page should be without Status for each Task card)
- [x] User Task page / Tabs Worked & Assigned - buttons 'Start' should be in 3rd column with Time (not where the progress bar → see Figma design)
- [x] User Task page / Tabs Worked & Assigned & Unassigned - make wider Estimate & Time columns
- [x] User Task page / Tabs Worked & Assigned & Unassigned - if the user refreshes the page it should remember Tab was selected the last time (now it always refers to Worked)
- [x] ~~Team Settings - when a user creates a custom Status, Priority, Size should select a custom icon if not selected in the purpose~~
- [x] Team page - Labels are displayed on user cards, if user reload whole page (but need to be checked code one more time, there is some delay or something wrong)
- [x] If Time tracking is disabled for a user - should not see 'own Task card' on Team page and 'own Task page'
- [x] the Size of Status, Labels, Priorities are still bigger than task titles etc [(PR)](https://github.com/ever-co/ever-gauzy-teams/pull/780)
- [x] When the manager deletes only team -> should see the main page without team + button 'Create team' 
- [x] When the manager deletes team and press 'log out' should see the main landing page (Now nothing)
- [x] When the manager deletes own account -> should see the main landing page (Now nothing)
- [x] When the manager tries to 'Remove account' from all teams (within one tenant/workspace) -> should be able to remove from all teams where the user is not a manager [(PR)](https://github.com/ever-co/ever-gauzy-teams/pull/805)
- [x] When a user has a few invites to join teams and clicks on ‘Reject’ on one of them, it should close only exactly that notification user ‘Rejected’ (Now it closes all notifications)
- [x] Team page / User card - Progress Bar does not display progress
- [x] User Task page / Task card - Progress Bar does not display progress [(PR)](https://github.com/ever-co/ever-gauzy-teams/pull/808)
- [x] Team page / User card - the manager should be able to estimate own and other’s tasks on the user card by clicking the ‘icon’ edit (it used to work)
- [x] Team page / User card / ( … ) Edit Task - user should be able to see the cursor at the end of the Title text inside the input AND the whole title should be selected (it used to work)
[...]

PR #777, opened 2023-04-12
fix: #734 Removed Status title in Team page

badalkhatri0924:
### Task - #734 

- [x]  Team page - on cards Titles - remove title 'Status' and Keep only 'Name' title and rest titles that we had


evereq:
@badalkhatri0924 can you show picture before and after changes, I am not sure why we remove status as it's very important element for any ticket.

badalkhatri0924:
> @badalkhatri0924 can you show picture before and after changes, I am not sure why we remove status as it's very important element for any ticket.

@evereq 
image
image

PR #814, opened 2023-04-26
fix: #734 Hide user card if time tracking is disable for manager

badalkhatri0924:
### Task - https://github.com/ever-co/ever-gauzy-teams/issues/734

- [ ] Team page - if a user (manager) disabled Time tracking, User task card should NOT be displayed on Team page

### DIFF
--- a/apps/web/lib/features/team-members.tsx
+++ b/apps/web/lib/features/team-members.tsx
@@ -16,7 +16,13 @@ import InviteUserTeamCardSkeleton from '@components/shared/skeleton/InviteTeamCa
 import { UserCard } from '@components/shared/skeleton/TeamPageSkeleton';
 // import { useEffect } from 'react';
 
-export function TeamMembers({ publicTeam = false }: { publicTeam?: boolean }) {
+export function TeamMembers({
+	publicTeam = false,
+	isTrackingEnabled = true,
+}: {
+	publicTeam?: boolean;
+	isTrackingEnabled?: boolean;
+}) {
 	const { isTeamManager, user } = useAuthenticateUser();
 	const { activeTeam, teamsFetching } = useOrganizationTeams();
 	const { teamInvitations } = useTeamInvitations();
@@ -47,7 +53,7 @@ export function TeamMembers({ publicTeam = false }: { publicTeam?: boolean }) {
 		<ul className="mt-7">
 			{/* Current authenticated user members */}
 			<Transition
-				show={!!currentUser}
+				show={!!currentUser && isTrackingEnabled}
 				enter="transition-opacity duration-75"
 				enterFrom="opacity-0"
 				enterTo="opacity-100"
--- a/apps/web/pages/index.tsx
+++ b/apps/web/pages/index.tsx
@@ -42,7 +42,11 @@ function MainPage() {
 			</MainHeader>
 
 			<Container className="mb-10">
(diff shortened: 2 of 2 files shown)


evereq:
@badalkhatri0924 / @KostyaEver I am not sure understand this PR / task. Why we should hide the employee card if time tracking is disabled? I think we still can see what employee was worked on (task assigned), estimate etc, it's just that employee can't track time using our time tracking tools, but s/he still can report tasks, time spent etc. Can you explain better what exactly we hiding in this PR and why?

evereq:
Adding more to my above comment: if employee time tracking is disabled, we should basically only hide "Start / stop and timer clock" for such employee, but all other features should be there and s/he can keep using Gauzy Teams platform with what we call manual time tracking or even without it (e.g. in Gauzy we can submit manually work time periods, we don't have that features yet in Gauzy Teams, but at the same time the goal right now is still to allow such employees to select tasks assigned, change tasks statuses, estimates, review progress of other team members, etc.

badalkhatri0924:
> @badalkhatri0924 / @KostyaEver I am not sure understand this PR / task. Why we should hide the employee card if time tracking is disabled? I think we still can see what employee was worked on (task assigned), estimate etc, it's just that employee can't track time using our time tracking tools, but s/he still can report tasks, time spent etc. Can you explain better what exactly we hiding in this PR and why?

Agree @evereq I think we should not hide user card if tracking is disable.

KostyaEver:
@evereq @badalkhatri0924 
I did not know about new ideas @evereq 
Last time when we discussed it. I mentioned that disabling time tracking and no need to see the user card, but the user (manager) still can create/manage/assign/unassign tasks to/from teammates. Now if you want to do like described above ok. it makes sense in the case of manual time adding.

evereq:
@KostyaEver well, why to "hide" person from the team if s/he can't track time? It still very useful to see person assigned tasks, even to see that s/he part of the team? There is nothing new in this and let's try to make sure we only do requirements that makes logical sense. There is no sense to hide team member if only time tracking disabled for her/him, I can't imagine single use case why we would want to do it that way. So let's fix this PR to only hide "time tracking" related things when employee time tracking is disabled, but keep everything else displayed. @badalkhatri0924

evereq:
Also, maybe we confuse what is "user card", please add screenshot below and explain what you mean by it @badalkhatri0924 / @KostyaEver . In my understanding it's a small card (row) that display info about user task, status (working / not working), estimate, even "..." actions menu and it's ALL makes sense to show even if person never tracking time with timers...

badalkhatri0924:
> Also, maybe we confuse what is "user card", please add screenshot below and explain what you mean by it @badalkhatri0924 / @KostyaEver . In my understanding it's a small card (row) that display info about user task, status (working / not working), estimate, even "..." actions menu and it's ALL makes sense to show even if person never tracking time with timers...

This card @evereq 
image

evereq:
Yes, so we need to display that card @badalkhatri0924 / @KostyaEver , no matter if time tracking enabled or disabled

badalkhatri0924:
@evereq You can merge it now after reviewing it. Now I am showing user card in both cases timer enabled/disabled. And just removed timer UI is it is disabled.

You can take reference from Screenshots attached.

CC: @KostyaEver

evereq:
@badalkhatri0924 I don't see screenshot attached :)

Issue #785, opened 2023-04-14
Autocomplete Create new team


KostyaEver:
**Create New Team**  page
- The system should **validate** Team Names the user enters - if there is an email, should ask to enter a different name
- Let's **remove** the **autocomplete** feature for this page

**Login to The Team** page
- Let's move the **autocomplete** feature for emails from **Create New Team** page

PR #826, opened 2023-04-28
Feat/improve task filter

paradoxe35:
- Auto close task filter when empty value
- disable autocomplete Create new team #785
- fixed(#734): assigned manager can not Remove Team Owner


apps/web/lib/settings/icon-popover.tsx
```diff
@@ -78,23 +77,23 @@ const IconPopover = ({
...
-													<DocumentUploadIcon className="w-5 h-5" />
-													<input
-														type="file"
-														className="absolute inset-0 opacity-0"
-														accept="image/svg+xml"
-													/>
-												</div>
-											</Tooltip>
-										}
+										// trailingNode={
```
evereq, line 80: @paradoxe35 why below code is commented out? For what it was used and why we don't need it anymore?

paradoxe35: No, we don't need it.
  It's an input file to upload a custom icon for a task status, since that it's a bad idea, so I commented it out

evereq: hm, why it's a bad idea to upload a custom icon for a task status!?
CC: @KostyaEver

PR #946, opened 2023-06-14
Member Actions, Permission Popup

badalkhatri0924:
image

image


badalkhatri0924:
@evereq You can merge this PR as it is getting bigger. I've updated the design and made new page for permissions. API integrations is pending so I've hide the field for now.

image

PR #990, opened 2023-07-05
Role Permission dynamic APi integration

badalkhatri0924:
image

image


evereq:
@badalkhatri0924 so we can merge now or not yet?

badalkhatri0924:
> @badalkhatri0924 so we can merge now or not yet?

@evereq Yes we can merge this now, but the Change Role thing from Member settings is pending, Will take it later on.

PR #994, opened 2023-07-06
Feat/unassign task

desperado1802:
Added the unassign task functionaloty

- Added it in the same popup with the assign member to task functionality
- Members already assigned to task are displayed first and have trashIcon to indicate that by pressing they can be removed

[screen-capture (1).webm](https://github.com/ever-co/ever-gauzy-teams/assets/124465103/0e77fce3-7013-4f5f-9b13-f7a17854e23b)

### DIFF
--- a/apps/web/components/pages/task/details-section/blocks/task-main-info.tsx
+++ b/apps/web/components/pages/task/details-section/blocks/task-main-info.tsx
@@ -266,14 +266,14 @@ const ManageMembersPopover = (
 						leaveTo="opacity-0 translate-y-1"
 					>
 						<Popover.Panel
-							className="z-10 absolute right-0 bg-white dark:bg-[#202023] rounded-2xl w-[9.5rem] flex flex-col pl-5 pr-5 pt-2 pb-2 mt-10 mr-10"
+							className="z-10 absolute right-0 bg-white dark:bg-[#202023] rounded-2xl min-w-[9.5rem] flex flex-col pl-4 pr-4 pt-2 pb-2 mt-10 mr-10"
 							style={{ boxShadow: 'rgba(0, 0, 0, 0.12) -24px 17px 49px' }}
 						>
 							{({ close }) => (
 								<div className="">
 									{assignedTaskMembers.slice(1).map((member, index) => (
 										<div
-											className="flex items-center justify-between h-8 w-auto hover:cursor-pointer hover:brightness-95 dark:hover:brightness-105"
+											className="flex items-center justify-between gap-1 h-8 w-auto hover:cursor-pointer hover:brightness-95 dark:hover:brightness-105 border-b mt-1"
 											onClick={() => {
 												setMember(member);
 												setMemberToRemove(true);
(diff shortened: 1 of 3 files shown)


evereq:
@desperado1802 hm, when you click "+ Add new member" you actually not only can add members but also can remove right? If so, why it's called "Add"? :) The way how you implemented should be called "Manage Assignees".
The alternative would be to have only the ability to ADD members that way, but to remove members you just put tiny "remove" icon near to each member that is already displayed (which might be a bit annoying honestly, so not sure I would do it that way)

evereq:
Few more questions:
there are tons of commented code there which actually looks like code that makes some sense too, why are we removing/rewriting all those codes? If it's really not needed, why do we keep it? Please describe how it worked before that PR so I understand why so much code was removed and not needed anymore...

badalkhatri0924:
> @desperado1802 hm, when you click "+ Add new member" you actually not only can add members but also can remove right? If so, why it's called "Add"? :) The way how you implemented should be called "Manage Assignees". The alternative would be to have only the ability to ADD members that way, but to remove members you just put tiny "remove" icon near to each member that is already displayed (which might be a bit annoying honestly, so not sure I would do it that way)

@evereq I think we can keep "Manage Assignees" and the rest implementation is good to go.

badalkhatri0924:
@desperado1802 Border under each name is fine, but it should not be for the last record. Let us just remove border-bottom for all the members, we are not using that UI in any other place.

evereq:
@desperado1802 what about https://github.com/ever-co/ever-gauzy-teams/pull/994#issuecomment-1624316266? Basically it's best in such tickets to show how it was before and after your changes (screenshots / loom), especially if so much code is commented out, we need to understand what those code was doing and why it's removed (if refactored it's good, but if by incident you removed it?)

desperado1802:
> @desperado1802 what about [#994 (comment)](https://github.com/ever-co/ever-gauzy-teams/pull/994#issuecomment-1624316266)? Basically it's best in such tickets to show how it was before and after your changes (screenshots / loom), especially if so much code is commented out, we need to understand what those code was doing and why it's removed (if refactored it's good, but if by incident you removed it?)

@evereq Some of the commented code was mine, as first I added second button to remove members from a given task. And the other half was the functionality to add member. In order to be able to assign and remove from one button I created a new component by following similar functionality from the commented code. Left it as commented untill the PR was reviewed(If I'm asked to turn it back as it was default to be able to do it faster). Now I removed the commented code. From now on I will screen record before any changes are implemented to show the difference, however basically the previous code was assigning members to a given task

badalkhatri0924:
@evereq you merged too early!

The creator of the task is missing in the Dropdown.

Anyways @desperado1802 please take the latest pull of develop and make those changes. I'll review again in that branch.

evereq:
Ops @badalkhatri0924 , I notice that too haha, but was not sure if we should allow to always keep the creator in the "Assignees" list or the opposite to NOT having the creator there at all... 
I see that we have a different field for Creator already...
So I think we should allow to add/remove Task Creator into/from assignees too... (and by default we should NOT add creator into list of assignees, i.e. when task created, assignees list should be empty).
Also, yes @desperado1802 please create another branch based on develop and work on it, do NOT reuse this branch from this PR please because I used method "Squash" to merge your changes. 

Thanks guys :)

PR #993, opened 2023-07-06
#520 task page side panel status block

paradoxe35:
- fixed #520 

User can create new status in the dropdown menu

PR #1003, opened 2023-07-08
506 task page side panel start due dates

### LINKED ISSUE #506: Task page - Side panel | Start & Due dates
An Issue should have approximate start and due dates

- Start (when a team should start working on it)
- Due (when a team should stop working on it)

Screenshot at Feb 10 13-06-36

Note: it will remind users that start/due dates are upcoming

paradoxe35:
- Finalize #993
- Fixed #506


badalkhatri0924:
@paradoxe35 

It is taking time by default. Should we keep it or not?

@evereq Should we take/show time too for tasks? It is enough to show the Date only.

Also if we are showing the time, we should also have a picker for the same. Let's know if @paradoxe35 needs to change it or not.

@paradoxe35 It will be good If we can have some styling based on our theme. This looks so vanilla!

image

paradoxe35:
Okay.
Let’s do it on the next PR @badalkhatri0924 

Or we can create a separate ticket for that (styling)

badalkhatri0924 (approved):

evereq:
@paradoxe35 / @badalkhatri0924 
1) I think "Due Date" should really be just a date without time :) because usually I deadlines set in dates :)))
2) what calendar we are using that it looks so vanilla? We are using some components lib already right (Chakra / Tailwind / whatever, not recall), they don't have well styles calendar? (both dark / light theme) That's weird to me....

badalkhatri0924:
> @paradoxe35 / @badalkhatri0924
> 
> 1. I think "Due Date" should really be just a date without time :) because usually I deadlines set in dates :)))
> 2. what calendar we are using that it looks so vanilla? We are using some components lib already right (Chakra / Tailwind / whatever, not recall), they don't have well styles calendar? (both dark / light theme) That's weird to me....

Well, we don't have any component library.

evereq:
@badalkhatri0924 / @paradoxe35 well, we use 'headlessui', 'tailwindcss' etc (not really component libs, but you get what I mean :D). I found https://github.com/onesine/react-tailwindcss-datepicker, maybe we can use that one to select single date too? If so, we may use it as it's already styled and it can be used to select ranges of dates later (we use similar already in Gauzy UI in Angular!). However, we will also probably need to migrate from momentjs to dayjs for frontend (needs to check how heavy it will be). But in any case, I don't want to spend too much time on styling old react datepicker for Tailwind (unless we will find someone already did it for both dark and light)

evereq:
@paradoxe35 I will merge this one for now, but please update me on Slack with results of your checks according to my comment above https://github.com/ever-co/ever-gauzy-teams/pull/1003#issuecomment-1627469825

paradoxe35:
> @paradoxe35 I will merge this one for now, but please update me on Slack with results of your checks according to my comment above [#1003 (comment)](https://github.com/ever-co/ever-gauzy-teams/pull/1003#issuecomment-1627469825)

Okay @evereq

PR #1017, opened 2023-07-12
Feat/#1004 task description serialize html

desperado1802:
Added serializer

Now the editor content is getting saved as html in the DB

- Used the slate-serializers library
- Made it so the editor accepts html, and also slate JSON object as value, in order to prevent errors from created untill now descriptions
- Configured it as it wasn't adding text aligns, block elements to the DB html
- Some UI fixes: added predefined height to the editor and made it scrollable, fixed h1 and h2 font size, placeholder margin, line margins, 
[screen-capture (2).webm](https://github.com/ever-co/ever-gauzy-teams/assets/124465103/0a651341-1dcf-4905-bc0f-279ee2bae931)

### DIFF
--- a/apps/web/components/pages/task/description-block/editor-components/BlockButton.tsx
+++ b/apps/web/components/pages/task/description-block/editor-components/BlockButton.tsx
@@ -9,7 +9,7 @@ interface IMarkButtonProps {
 }
 
 const TEXT_ALIGN_TYPES = ['left', 'center', 'right', 'justify'];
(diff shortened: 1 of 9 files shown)


paradoxe35 (approved):
LGTM @desperado1802 

We can also consider supporting copy-paste keeping the original text format.

https://www.slatejs.org/examples/paste-html
https://github.com/ianstormtaylor/slate/blob/main/site/examples/paste-html.tsx

Cc @evereq

Issue #1020, opened 2023-07-13
UI | Team Settings - Fix input text color when dark theme


paradoxe35:
<img width="790" alt="Screenshot 2023-07-13 at 12 28 56" src="https://github.com/ever-co/ever-teams/assets/35149259/eda104cf-d554-486c-b90c-97c029323969">
-----
<img width="833" alt="Screenshot 2023-07-13 at 12 29 02" src="https://github.com/ever-co/ever-teams/assets/35149259/ac3bb5b2-87ba-4b0a-8de7-dfc7e18cc465">

Issue #521, opened 2023-02-13
Task page - Linked Issues block


KostyaEver:
An Issue should have Linked Issue blocks (under the Description block), 
where are listed 
- Child Issues (of the current issue)
- Related Issues (of the current issue)

Users can add Children and Related Issues to the Current Issue by clicking '+' button which calls Pop with Issues

<img width="586" alt="Screenshot at Feb 13 12-20-15" src="https://user-images.githubusercontent.com/107203576/218448091-b7b34011-ac8f-489c-97e6-69f682916b5b.png">

main95:
Hi, @KostyaEver, i'm new in the project, can I take this issue?

evereq:
Hey @main95 , thank you for offering help with our project! This task actually depends on functionality of our Gauzy APIs, which I don't think is done at this stage, so this task can't proceed. If you have time to help our project(s) development we can discuss what task is a good to start etc. Let me know if it's interesting for you.

main95:
Hi @evereq, yes I'm interesting in help to yours projects, I'm finding some side projects in my web dev career, do you have some infos?

evereq:
@KostyaEver and me will check what tasks we can delegate @main95 , we will message you next week (having holidays now)

main95:
ok, thanks

badalkhatri0924:
@paradoxe35 

Child Issue/task related API changes are done.

In GET API you need to pass `children` and `parent` relation.

We have introduced new `parentId` field. you can add logic in update API to handle this task.

PR #1021, opened 2023-07-13
521 task page linked issues block

paradoxe35:
- #521 hooks and UI  
- Fix input text color when dark theme #1020
- Other minor fixes


apps/web/pages/task/[id].tsx
```diff
@@ -68,7 +68,7 @@ const TaskDetails = () => {
 							<RichTextEditor />
 							{/* <TaskDescriptionBlock /> */}
 							<IssueCard related={false} />
-							<IssueCard related={true} />
+							{/* <IssueCard related={true} /> */}
```
Comment (evereq) [inline apps/web/pages/task/[id].tsx:71]: @paradoxe35 that one with `related={true}` is not implemented yet or you forgot to uncomment?

Comment (paradoxe35) [reply apps/web/pages/task/[id].tsx:71]: Not yet implemented

Issue #510, opened 2023-02-10
Task page - Side panel | more details about issue


KostyaEver:
An Issue should have more details
- Date and time of the creation
- Date and time of the last update (status changes, edit task description etc)
- Date and time when was resolved (Selected Type of Status Done or Closed)

**Note:** the details should be placed under Side Panel

<img width="200" alt="Screenshot at Feb 10 15-02-08" src="https://user-images.githubusercontent.com/107203576/218110854-a31b0773-6f38-49a5-b41f-437ef25c8fdc.png">

PR #1033, opened 2023-07-17
created the task-properties bellows the task sidebar

desperado1802:
#510 Created a component that shows when the task is created, updated and resolved. 

The resolved date-time string currently is null, however functionality is added, when from DB date comes it will be reflected to the DOM

Before:
image

After:
image


badalkhatri0924:
@evereq we can merge this.

PR #1034, opened 2023-07-18
Feat/task editor checkbox

desperado1802:
Created the checkboxes for the taks description editor. They can we saved to DB as html and retrieved from there back as Slate object

https://github.com/ever-co/ever-teams/assets/124465103/1bdca48e-1e29-45f0-949b-f066ca36ddfc


badalkhatri0924:
@desperado1802 Is it possible to remove these checkboxes?

Basically, I am trying to delete empty 2 checkboxes using Backspace/Delete button, but it is not working.


image

evereq:
@desperado1802 also can we please fix merge issue?

apps/web/components/pages/task/description-block/editor-components/serializerConfigurations.ts
```diff
@@ -24,6 +25,22 @@ export const configHtmlToSlate: HtmlToSlateConfig = {
...
+				return {
+					type: 'checklist',
+					checked: checkbox && checkbox.attribs.checked === 'true',
+					children: [
+						//@ts-ignore
+						{ text: textNode && textNode.firstChild.data },
+					],
+				};
+			}
+			return undefined;
```
evereq, line 42: @desperado1802 seriously, return undefined? :)

desperado1802: It returns undefined when the condition is not met. If the div element from DB doesn't have class "checklist-container". That's the only type div element we have as editor html value anyway, for that reason it won't return undefined

PR #1065, opened 2023-07-24
Bug/team settings disable

desperado1802:
Added functionalities team settings fields to be disabled by default, and to be enabled, and dialogs to appear only when edit icon is pressed.

- All team settings fields are disabled by default
- Dialogs appear only when edit icon is pressed
- When pressed team name's edit icon field gets focused, when pressed outside of it gets disabled. Saves only when edit icon pressed again. I think we might add "check" icon or something to indicate user how to save team name
- Emiji field's delete icon wasn't working, added functionality

Before:
https://github.com/ever-co/ever-teams/assets/124465103/df74b457-d4d7-49d6-9a10-029d242ed2fe

After:
https://github.com/ever-co/ever-teams/assets/124465103/e8c995eb-1aab-4e12-bca8-6690c22e7395


badalkhatri0924:
@desperado1802 It is not looking good (UX) wise that the user has to click again on the same "Edit" icon. I think we should change the icon to a "Save" icon when the field is in edit mode.

Issue #1067, opened 2023-07-25
BUG: User detail page, Alignment of menu of task keeps changing on hover


badalkhatri0924:
[Ever-Teams-User-Page-Task-Menu-Alignment-Issue-On-Hover.webm](https://github.com/ever-co/ever-teams/assets/81486442/2d8f342c-3f4e-4764-8c46-4e0cead50f76)

Issue #1066, opened 2023-07-25
BUG: User details page, menu of task is automatically closed (even user do not click anywhere)


badalkhatri0924:
[Ever-Teams-User-Page-Menu-Auto-Hide.webm](https://github.com/ever-co/ever-teams/assets/81486442/96c7e37d-9cc7-478d-938e-21c80e7791b8)

Issue #1059, opened 2023-07-24
BUG: Not able to set Due date same as Start Date


badalkhatri0924:
[Ever-Teams-Due-Date.webm](https://github.com/ever-co/ever-teams/assets/81486442/3b82219b-3e86-450e-8547-5aa5fa71285d)

Issue #1058, opened 2023-07-24
BUG: Child issue can not be collapsed


badalkhatri0924:
<img width="934" alt="image" src="https://github.com/ever-co/ever-teams/assets/81486442/065eb9c4-cd08-4677-954c-761ae6cd25f2">

PR #1097, opened 2023-07-26
1059 bug not able to set due date same as start date

paradoxe35:
- #1059
- #1058
- #1057
- #1056
- #1067
- #1066


badalkhatri0924:
Start & Due dates are in the future, but the Days Remaining are being counted from today's date.

@evereq This is expected behavior or we should change it? Maybe if the start date is in the future we can hide the Days Remaining UI.

image

CC: @paradoxe35

badalkhatri0924 (approved):

evereq:
@badalkhatri0924  Hm, interesting question :))) Depends really on what does it mean "days remaining". Please ask @KostyaEver (possibly in Slack), maybe he can explain more about it. In my view, yes, it can be how many days remain from now to the Due date OR it can be how many days of WORK remain to complete it... i.e. related to estimate value, but in this case not clear why it's days? It can also be hours too... E.g. task was estimated 10 hours and I worked 3 hours on it, so the remaining is 7 hours, not days... 
Think let's discuss in Slack, I will merge PR for now anyway as is :)

PR #1146, opened 2023-08-02
1087 task detail page handle case to automatically select version

### LINKED ISSUE #1087: Task Detail page: Handle case to automatically select Version
- If there are no versions already created and the user creates a new one then it should be automatically selected
- If there are already >= 1 versions created and user creates new one then it should **not** be automatically selected

image

paradoxe35:
- Fix #1087


badalkhatri0924:
image

image
@paradoxe35 In the dropdown it is not showing `.` but actually it does have dots.

paradoxe35:
> image image @paradoxe35 In the dropdown it is not showing `.` but actually it does have dots.

@badalkhatri0924 this is another issue
Better fix it in a separate PR

badalkhatri0924 (approved):

PR #1163, opened 2023-08-04
Fix/task type dropdown enable

badalkhatri0924:
[Ever-Teams-Task-Type-Change.webm](https://github.com/ever-co/ever-teams/assets/81486442/f5a8c123-5b59-4932-90ea-60c2aacde897)

### As EPIC can't have Parent, removed "Add Parent" button:
image

### Story can have only EPIC as a Parent:
image

### When task has EPIC as a Parent:
image

### When task has Story as a Parent:
image


paradoxe35 (approved):
Looks Good @badalkhatri0924

PR #1199, opened 2023-08-08
added the circle icon to the dropdowns when they have no value

desperado1802:
Added circle icon when no value is selected in the dropdowns. Added to all dropdowns as in my opinion it looks more consistent

image

image


badalkhatri0924:
@desperado1802 Circle icon is bit larger than expected.

badalkhatri0924 (approved):

PR #1201, opened 2023-08-08
added save icon to the estimate input, user-card

desperado1802:
Added save icon to user card task estimations

https://github.com/ever-co/ever-teams/assets/124465103/2e4c17f5-e97b-4227-a234-ecd9697d8a5d


badalkhatri0924:
@desperado1802 

- When the user tries to edit the estimation, and h/she does not change values and directly clicks on the save icon, it is not working.
- Also not sure why clicking outside of input is not saving now(was working before)

badalkhatri0924 (approved):

PR #1203, opened 2023-08-08
Fix/Passcode Page Auth Cards Animation

paradoxe35:


badalkhatri0924:
@paradoxe35 left side borders are not as rounded as right side one.

paradoxe35:
The Image above is not the result (it’s the issue)
@badalkhatri0924

badalkhatri0924:
> The Image above is not the result (it’s the issue) @badalkhatri0924

Oh my bad, just checked. Looks good.

badalkhatri0924 (approved):

PR #1244, opened 2023-08-14
Updated Manager icon in user card

badalkhatri0924:
image
image


badalkhatri0924:
@evereq Here should we show "Admin", "Super Admin"... or just Manager?

evereq:
@badalkhatri0924 manager is good, but I would make the height of those tooltips might smaller... I.e. close to 1/2 really, now it's too big everywhere...

PR #1270, opened 2023-08-16
Feat/Related Task Issue

paradoxe35:
Screenshot 2023-08-16 at 14 34 00


evereq:
@Babajide777 can you review as priority please so we merge?

evereq:
One notice: in the screenshot, we can see that status "In Progress" did not fit, let's make those fields bigger so status "IN PROGRESS" can fit into them. @paradoxe35 
CC: @Babajide777

evereq:
@paradoxe35 yes, but for so important things as Status for the task, i would really prefer to make the combobox bigger width so it fits that specific common status "IN PROGRESS", especially as we see there is tons of empty space there... We don't need to make it much longer, just a bit more than now and it will fit text well!

paradoxe35:
You're right @evereq 

@badalkhatri0924  let's create a separate PR for this, if you agree

badalkhatri0924:
> You're right @evereq
> 
> @badalkhatri0924 let's create a separate PR for this, if you agree

Yes lets do it in separate PR.

apps/web/components/pages/task/IssueCard.tsx
```diff
@@ -105,33 +141,25 @@ function CreateLinkedTask({
...
-			(task.issueType === 'Epic'
-				? t.issueType !== 'Epic'
-				: task.issueType === 'Story'
-				? // If current task is Story then filter Epics and Stories from the list
-				  t.issueType !== 'Epic' && t.issueType !== 'Story'
-				: t.issueType === 'Bug' ||
-				  t.issueType === 'Task' ||
-				  t.issueType === null)
-	);
+			!['Epic', 'Story'].includes(t.issueType)
```
badalkhatri0924, line 160: @paradoxe35 this logic is breaking. Please refactor it.

If current task is EPIC then STORY, TASK, BUG can be children
If current task is  STORY then TASK, BUG can be children
If current task is TASK|BUG then both can be children.

apps/web/components/pages/task/IssueCard.tsx
```diff
@@ -1,43 +1,62 @@
...
 	const task = useRecoilValue(detailedTaskState);
 	const { tasks } = useTeamTasks();
 	const [hidden, setHidden] = useState(false);
 
+	const { actionType, actionTypeItems, onChange } = useActionType();
+
 	const linkedTasks = useMemo(() => {
-		return (
+		const issues =
 			task?.linkedIssues
```
badalkhatri0924, line 39: @paradoxe35 can you try to improve this logic? multiple .filter and .map will reduce the performance when we have more tasks.

apps/web/components/pages/task/ChildIssueCard.tsx
```diff
@@ -0,0 +1,157 @@
...
+		return (
+			t.id !== task.id &&
+			!childTasks.includes(t.id) &&
+			!['Epic', 'Story'].includes(t.issueType)
+		);
+	});
+
+	return (
+		<Modal isOpen={modal.isOpen} closeModal={modal.closeModal}>
+			<div className="w-[98%] md:w-[668px] relative">
```
badalkhatri0924, line 127: @paradoxe35 can we use `rem` instead of `px` here? Lets shift to `rem` in whole projects gradually for old code. and new code we use rem only.

badalkhatri0924 (changes requested):

badalkhatri0924:
image

@paradoxe35 plz update the divider the same as we have at other places. the white divider is not looking good.

badalkhatri0924 (approved):

PR #1278, opened 2023-08-21
Fixed Settings Screen toggle dark button

desperado1802:
IMG_0598
IMG_0599


apps/mobile/app/screens/Authenticated/SettingScreen/components/SingleInfo.tsx
```diff
@@ -40,11 +40,14 @@ const SingleInfo: FC<Props> = ({ title, value, onPress, onDetectTimezone }) => {
...
+				<TouchableOpacity
+					style={dark ? styles.toggleDark : styles.toggleLight}
+					onPress={() => (onPress ? onPress() : {})}
+				>
 					{dark ? (
-						<Image style={{}} source={require("../../../../../assets/icons/new/toogle-dark.png")} />
+						<Image style={{}} source={require("../../../../../assets/icons/new/toggle-dark.png")} />
 					) : (
-						<Image source={require("../../../../../assets/icons/new/toogle-light.png")} />
+						<Image source={require("../../../../../assets/icons/new/toggle-light.png")} />
```
badalkhatri0924, line 50: @desperado1802 I don't think we should use images for the toggle, we should implement actual toggle button logic, the same as we have in Web.

badalkhatri0924 (changes requested):

apps/mobile/app/screens/Authenticated/SettingScreen/components/SingleInfo.tsx
desperado1802: Used the same logic we use in the web app
https://github.com/ever-co/ever-teams/assets/124465103/83c54fcb-6ee0-4422-8259-15c453265cec

badalkhatri0924 (approved):

PR #1280, opened 2023-08-21
Feat/Action type dropdown

paradoxe35:
Screenshot 2023-08-22 at 15 27 05


badalkhatri0924 (approved):

badalkhatri0924:
image


@evereq We can merge this PR. but I feel dropdown size looks small (though we have implemented same as figma).

If require we can increase the size of overall UI in separate PR.

PR #1302, opened 2023-08-29
Jitsi Integration front

paradoxe35:
https://www.loom.com/share/a9bc071939b04682b49ab8f7d7d5a841


paradoxe35:
@Babajide777 
You can review this PR, for others modifications
I'll create a dedicated PR

badalkhatri0924 (approved):

apps/web/lib/app/authenticator.tsx
```diff
@@ -55,17 +63,17 @@ export function withAuthentication(
...
 		return (
 			<>
-				{!user || loading ? (
+				{(!user || loading) && showPageSkeleton ? (
 					<TeamPageSkeleton />
 				) : (
 					<>
 						<Component {...props} />
-						<BackdropLoader
+						{/* <BackdropLoader
```
evereq, line 71: @paradoxe35 why this code commented!?

evereq:
@paradoxe35 let's keep building integration in this branch, it's too early to merge it.
Please merge periodically `develop` branch to this one so you keep it up to date with other developments, but we need to implement Auth and I also don't see how it works for real meetings from your video and basically nothing yet working (e.g. can't save whiteboard / can't share it etc)

apps/web/lib/app/authenticator.tsx
paradoxe35: We no longer use this loading component, we have replaced it with a skeleton component.
It created an empty DOM element that overly other elements and made them impossible to click.

evereq: ok, so maybe we can remove that code completely in such case? :)

paradoxe35: We can remove it

paradoxe35:
Jitsi SSO, Here is demo:
https://www.loom.com/share/897c5a7e0b6d482b9c24d21ff7ac947e

So basically, it's using Jitsi JWT authentication.

First, we define a secret:

This secret will be used to verify the JWT token on the Jitsi server.
We will also use the same key to encrypt the JWT token from the Ever Team Next.js server API. This key must remain private and not accessible on the browser side.
This is the body structure of the JWT token:

```json
{
    "context": {
        "user": {
            "avatar": "https:/gravatar.com/avatar/abc123",
            "name": "John Doe",
            "email": "jdoe@example.com",
            "id": "abcd:a1b2c3-d4e5f6-0abc1-23de-abcdef01fedcba"
        }
    },
    "aud": "jitsi",
    "iss": "ever_teams",
    "sub": "meet.jit.si",
    "room": "*"
}
```

https://github.com/ever-co/ever-teams/blob/0fe992dd23bb921bfbf0d09319c7a2be44a667c7/apps/web/pages/api/auth/jitsi/jwt.ts#L28-L45


@badalkhatri0924 / @evereq

paradoxe35:
@badalkhatri0924 

To review this you need your own self hosted instance of Jitsi

evereq:
Let's keep conversations in Issues (e.g. https://github.com/ever-co/ever-teams/issues/1334, I just created it), not in PRs @paradoxe35 . I will reply to your question there.

apps/web/middleware.ts
```diff
@@ -16,7 +16,9 @@ export const config = {
 		'/auth/(.*)',
 		'/profile/:path*',
 		'/settings/(.*)',
-		'/task/(.*)',
+		'/task(.*)',
+		'/call(.*)',
```
evereq, line 20: @paradoxe35 I think that should be now "meet", not "call"?

apps/web/lib/features/integrations/jitsi.tsx
```diff
@@ -0,0 +1,42 @@
...
+		.replaceAll(' ', '');
+
+	return (
+		<JitsiMeeting
+			domain={JITSI_DOMAIN}
+			roomName={'EverTeam' + roomName}
+			configOverwrite={{
+				startWithAudioMuted: true,
+				startWithVideoMuted: true,
+				disableModeratorIndicator: false,
```
evereq, line 22: @paradoxe35 any reason why moderator indicator disabled?
Also, `startScreenSharing`, does it mean that when video started it automatically start screen sharing if that option enabled (if so, let's keep it false).

apps/web/app/constants.ts
```diff
@@ -8,8 +8,11 @@ export const PROTECTED_APP_URL_PATHS: RegExp[] = [
 	/^(\/profile(\/)?)(.*)$/,
 	/^(\/settings(\/)?)(.*)$/,
 	/^(\/task(\/)?)(.*)$/,
+	/^(\/call(\/)?)(.*)$/,
```
evereq, line 11: @paradoxe35 I think here should be "meet" too instead of "call"?

apps/web/middleware.ts
evereq: I fixed that myself @paradoxe35 , let me know if I did something wrong :D

apps/web/app/constants.ts
evereq: I fixed that myself @paradoxe35 , let me know if I did something wrong :D

paradoxe35: Yes @evereq 
I should be meet

Issue #1409, opened 2023-09-21
[Bug] Public Page not showing Accurate Data


badalkhatri0924:
<img width="1487" alt="image" src="https://github.com/ever-co/ever-teams/assets/81486442/7d49cc76-78db-495e-bb6e-768bcafb0b01">

PR #1412, opened 2023-09-21
fix: Current Timer Status

badalkhatri0924:
https://github.com/ever-co/ever-teams/issues/1409

### DIFF
--- a/apps/web/lib/features/team/user-team-card/user-info.tsx
+++ b/apps/web/lib/features/team/user-team-card/user-info.tsx
@@ -75,9 +75,9 @@ export function UserInfo({ className, memberInfo, publicTeam = false }: Props) {
 									? 'pause'
 									: !member?.employee?.isActive && !publicTeam
 									? 'suspended'
-									: member?.employee?.isOnline &&
-									  member?.timerStatus !== 'running'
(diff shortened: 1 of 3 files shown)


apps/web/lib/features/team/user-team-card/user-info.tsx
```diff
@@ -75,9 +75,9 @@ export function UserInfo({ className, memberInfo, publicTeam = false }: Props) {
 									? 'pause'
 									: !member?.employee?.isActive && !publicTeam
 									? 'suspended'
-									: member?.employee?.isOnline &&
-									  member?.timerStatus !== 'running'
-									? 'online'
+									: member?.employee?.isOnline
+									? //  && member?.timerStatus !== 'running'
```
evereq, line 79: @badalkhatri0924 I don't get it? Why commenting it out will make it more accurate?

badalkhatri0924: @evereq when timer is running we are getting isOnline true.

Ideally timeStatus should return all the cases. Like Online, Online_and_tracking time etc. That I think we will do later on when we have real time.

PR #1420, opened 2023-09-22
Feat/#1419 collaboration flow

badalkhatri0924:


paradoxe35 (approved):
@badalkhatri0924 
Well done 👏

Cc @evereq

paradoxe35 (commented):
Screenshot 2023-09-25 at 23 00 28

Maybe one suggestion @badalkhatri0924 
What if we just use Icon instead of Icon + Text (Collaboration butto)

evereq:
> Screenshot 2023-09-25 at 23 00 28
> Maybe one suggestion @badalkhatri0924 What if we just use Icon instead of Icon + Text (Collaboration butto)

No, I think text on that button required to make it more visible that it's CTA

PR #1434, opened 2023-09-25
Imrpove/tracking disabled disabling timer in profile

desperado1802:
Added functionality, when tracking is disabled the Start Timer buttons in the Profile Screen do be disabled

https://github.com/ever-co/ever-teams/assets/124465103/889a4e69-9019-4081-b5ea-ec5de481da09


badalkhatri0924:
@desperado1802 Lets remove timer UI instead of reducing opacity, we do same in WEB.

badalkhatri0924 (approved):

PR #1449, opened 2023-09-26
Improve/sentry integration mobile

desperado1802:
Integrated Sentry to the Mobile app


apps/mobile/app/utils/crashReporting.ts
```diff
@@ -23,46 +23,45 @@
...
-  // Sentry.init({
-  //   dsn: "YOUR DSN HERE",
-  //   enableInExpoDevelopment: true,
-  //   debug: true, // If `true`, Sentry will try to print out useful debugging information if something goes wrong with sending the event. Set it to `false` in production
-  // })
-  // Bugsnag.start("YOUR API KEY")
+	Sentry.init({
+		dsn: process.env.EXPO_PUBLIC_SENTRY_DSN,
+		enableInExpoDevelopment: true,
+		debug: false, // If `true`, Sentry will try to print out useful debugging information if something goes wrong with sending the event. Set it to `false` in production
```
evereq, line 29: @desperado1802 we should detect here if it's production or dev/stage envs etc. Please create GitHub issue to improve that, but don't work on it yet, it will require some more things from me to be done.

evereq: Oh, actually @desperado1802 I notice we have already var __DEV__ (search code!) and you can use it to detect env (DEV or not dev). So debug should be enabled on DEV and disabled in any other env. Please fix that in that PR, no need to create GitHub issue!

apps/mobile/app/utils/crashReporting.ts
```diff
@@ -23,46 +23,45 @@
...
+		const message = error.message || "Unknown"
+		console.error(error)
+		console.log(message, type)
+		console.tron.log(error)
+	} else {
+		// In production, utilize crash reporting service of choice below:
+		// RN
+		// Sentry.captureException(error)
+		// Expo
+		// Sentry.Native.captureException(error)
```
evereq, line 63: @desperado1802 why you did not uncomment that line @desperado1802 ??? I.e. `Sentry.Native.captureException(error)`? That should work well I think!

desperado1802: Sure, I will uncomment it in the next push. Wanted to be sure that the configuration so far is implemented as it should. Also as I'm not able to test it on Expo Go(It doesn't have Sentry SDK/ Native events)

desperado1802: Did the changes you asked me to do @evereq. Did some documentation GitHub issues reading yesterday. There it's said that native events are not available for Expo Go. However in production and eas build should work. 

some of the references:
https://github.com/expo/sentry-expo/issues/361
https://docs.expo.dev/guides/using-sentry/

image

apps/mobile/app.json
```diff
@@ -76,6 +77,17 @@
 				"projectId": "9f38f768-16f6-4c47-b354-9745ebda1335"
 			}
 		},
+		"hooks": {
+			"postPublish": [
+				{
+					"file": "sentry-expo/upload-sourcemaps",
+					"config": {
+						"organization": "SENTRY_ORG",
```
evereq, line 85: @desperado1802 will it substitute those vars for SENTRY_ORG and SENTRY_PROJECT here from env vars or we have to hardcode or what? How will it work? In my understanding we should put there correct values so it can upload sourcemaps for stack traces etc to Sentry.io!

apps/mobile/app/utils/crashReporting.ts
```diff
@@ -23,46 +23,45 @@
  *  This is where you put your crash reporting service initialization code to call in `./app/app.tsx`
  */
 export const initCrashReporting = () => {
-  // Sentry.init({
-  //   dsn: "YOUR DSN HERE",
-  //   enableInExpoDevelopment: true,
-  //   debug: true, // If `true`, Sentry will try to print out useful debugging information if something goes wrong with sending the event. Set it to `false` in production
-  // })
-  // Bugsnag.start("YOUR API KEY")
+	Sentry.init({
```
evereq, line 26: Please add to the code above check if `process.env.EXPO_PUBLIC_SENTRY_DSN` is defined and only call .init method if it is. Otherwise skip all initialization/usage of Sentry! I.e. if we did not configure it, it should not cause app crash etc.

apps/mobile/app/utils/crashReporting.ts
```diff
@@ -23,46 +23,45 @@
...
+		const message = error.message || "Unknown"
+		console.error(error)
+		console.log(message, type)
+		console.tron.log(error)
+	} else {
+		// In production, utilize crash reporting service of choice below:
+		// RN
+		// Sentry.captureException(error)
+		// Expo
+		Sentry.Native.captureException(error)
```
evereq, line 63: Same, add here  check if `process.env.EXPO_PUBLIC_SENTRY_DSN` is defined and only call that method if yes

desperado1802:
@evereq Did the changes to env's as we discussed. Also tested react-native-dotenv and Expo's env, both working fine

Issue #1482, opened 2023-09-29
[Improvement] Labels Tags and Popup in Profile Section's Create Task Modal


desperado1802:
Current Implementation:
https://github.com/ever-co/ever-teams/assets/124465103/1f3aa698-bf52-4fc5-ba25-6df6f3d2d459

Needs to be same as in Timer Section labels tags, and popup, functionality and UI wise

PR #1483, opened 2023-09-29
Improve/create task labels profile section

desperado1802:
#1482
Added the same component that we use in Timer Section labels, with little modification to make it reusable for the case
https://github.com/ever-co/ever-teams/assets/124465103/6557bcf5-d54e-491c-954e-b7ede0fd9ba0


badalkhatri0924:
@desperado1802 Can we improve UX.

As per current implementation user has to open popup multiple times if he/she wants to select multiple tags. Ideally user should be able to select multiple tags and lets have Save button and on that Save button we will do API calling/state update etc...

badalkhatri0924 (approved):

PR #1489, opened 2023-10-03
updated the api url in env so both icons and production server can run

desperado1802:


apps/mobile/app/config/config.prod.ts
```diff
@@ -9,5 +9,5 @@
 import { GAUZY_API_URL } from "@env"
 
 export default {
-	API_URL: GAUZY_API_URL,
+	API_URL: `${GAUZY_API_URL}/api`,
```
badalkhatri0924, line 12: @desperado1802 plz do same for config.dev.ts.

badalkhatri0924 (changes requested):

badalkhatri0924 (approved):

Issue #1494, opened 2023-10-04
[BUG] Mobile App and Web timer sync


desperado1802:
When the timer is started from Web App, the Mobile App's timer doesn't sync(start running).The opposite works(starting from Mobile, starts the Web timer)

Starting the Timer in Web:
https://github.com/ever-co/ever-teams/assets/124465103/c1fcbf88-2f95-45ce-a8cb-9e5560129cad

Doesn't start the timer in Mobile:
https://github.com/ever-co/ever-teams/assets/124465103/d01ea538-4884-4ee9-a8d4-176e01861489

Later on we start the timer on Mobile, adn the Web timer automatically syncs. The opposite should also work

PR #1495, opened 2023-10-04
Bug/timer sync mobile

desperado1802:
#1494
Synced the timers when it's started from Web side, the opposite was working already. Also fixed API calls. Timer status wasn't fetching constantly to check if timer is running, now does every 5 sec

We start the Web App timer:
https://github.com/ever-co/ever-teams/assets/124465103/185b5d4f-7fbb-42b3-80b2-6794d20d7374

Mobile App syncs(starts too):
https://github.com/ever-co/ever-teams/assets/124465103/947f1a4f-df4a-4b1d-b489-19257e16fb83

### DIFF
--- a/apps/mobile/app/services/client/queries/timer/timer.ts
+++ b/apps/mobile/app/services/client/queries/timer/timer.ts
@@ -7,11 +7,6 @@ type IGetTimerStatusParams = ITimerTimeslotParams & { authToken: string }
 const fetchTimerStatus = async (params: IGetTimerStatusParams) => {
 	const { tenantId, organizationId, logType, authToken, employeeId } = params
 
-	const { data } = await getTimerStatusRequest(
-		{ source: "MOBILE", tenantId, organizationId },
-		authToken,
-	)
-
 	await syncTimeSlotRequest(
 		{
 			tenantId,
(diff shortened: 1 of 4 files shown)


apps/mobile/app/services/client/queries/timer/timer.ts
```diff
@@ -4,34 +4,36 @@ import { ITimerTimeslotParams } from "../../../interfaces/ITimer"
...
-	const { data } = await getTimerStatusRequest(
-		{ source: "MOBILE", tenantId, organizationId },
-		authToken,
-	)
+	if (isTimerRunning) {
+		await syncTimeSlotRequest(
+			{
+				tenantId,
+				organizationId,
+				source: "MOBILE",
```
badalkhatri0924, line 15: @desperado1802 I think this is wrong. 

We have hard coded MOBILE source sync, we have to make it dynamic.

For Example:
If user started timer in WEB or DESKTOP or at other source we have to pass that value in source (If we don't find that value then only we have to pass MOBILE).


Take Reference from:
https://github.com/ever-co/ever-teams/blob/develop/apps/web/app/hooks/features/useTimer.ts#L238


So basically we have to at which source current timer is running, and we have to pass that value.

badalkhatri0924 (changes requested):

badalkhatri0924 (approved):

Issue #1506, opened 2023-10-05
[BUG] Teams Screen > Task Card is not showing correct state of timer


desperado1802:
The Task Card is showing only Timer Runing(green) and Timer Idle(red). Should also show Timer Paused(yellow) with appropriate icon

![image](https://github.com/ever-co/ever-teams/assets/124465103/448f6e36-9746-404d-a6a2-d9421a9195ed)

As in Web App:
![image](https://github.com/ever-co/ever-teams/assets/124465103/8e158717-94be-4cad-a4bd-100f20cc9911)

PR #1508, opened 2023-10-05
Bug/teams screen cards timer state

desperado1802:
#1506
Updated the logic so it works similar to Web app. The User Cards show the pause state when timer is started in another device and stopped, or when it's started in Mobile app and stopped.

In the video first time the timer is started from the mobile, and that causes the state to be "Pause". The second time it's stared from web, when the timer is running the state becomes "Online", and after stopping becomes "Pause" again. The same behavior is implemented to Web App.

https://github.com/ever-co/ever-teams/assets/124465103/c44e921b-fb90-4908-99af-81767ae03b0a


apps/mobile/app/screens/Authenticated/TeamScreen/components/ListCardItem.tsx
```diff
@@ -148,13 +151,23 @@ const ListCardItem: React.FC<Props> = observer((props) => {
...
-					? "#9FDAB7"
-					: !props.member?.totalTodayTasks?.length
-					? "#F1A2A2"
-					: "#EBC386",
+				backgroundColor:
+					!timerStatus?.running &&
+					timerStatus?.lastLog &&
+					timerStatus?.lastLog?.startedAt &&
+					moment().diff(moment(timerStatus?.lastLog?.startedAt), "hours") < 24 &&
+					(timerStatus?.lastLog?.source !== "MOBILE" || props.member?.employee?.isOnline)
```
badalkhatri0924, line 159: @desperado1802 I don't get this line:

If member is isOnline at some other source then we will saw `Pause`?

badalkhatri0924: Also @desperado1802 If you see in video, not only current member's but all other member's status is shown "Paused".

badalkhatri0924 (approved):

PR #1553, opened 2023-10-13
Improve board menu

paradoxe35:
I removed the Excalidraw links

image

image


evereq:
@paradoxe35 why no screenshot / loom? :)

PR #1648, opened 2023-10-25
Feat/table view for issues

YehualashetGit:
* create data table ui component
    * create re usable table component and used them for Data table UI component
    * refactor team members componen so that it can show card and table component
    * add a swtich button to change issue view

https://www.veed.io/view/f1ebbac3-64a1-4423-8e96-05685b3ea2b6?panel=share


badalkhatri0924:
image



@YehualashetGit fix build issues.

badalkhatri0924:
@YehualashetGit why needed to force push?

YehualashetGit:
> @YehualashetGit why needed to force push?

It doesn't allow me to push after  merging the conflict

badalkhatri0924:
> > @YehualashetGit why needed to force push?
> 
> It doesn't allow me to push after merging the conflict

@evereq what do you suggest here?

badalkhatri0924:
image
Also @YehualashetGit it is not loading data for me.

YehualashetGit:
> image Also @YehualashetGit it is not loading data for me.

wait let me finalize it ,it's still under Draft PR, I will make it ready for return teamMembersView;
 once I'm done, here I was doing the filter by   user id ,

apps/web/app/constants.ts
```diff
@@ -116,3 +116,9 @@ export const APPLICATION_LANGUAGES_CODE = [
 	'ru',
 	'es'
 ];
+
+export enum KanbanView {
```
evereq, line 120: Maybe let's rename this to "IssuesView"?

apps/web/lib/features/team/invite/user-invite-card.tsx
```diff
@@ -225,9 +225,9 @@ export function InviteUserTeamCard({
 					<DraggerIcon className="fill-[#CCCCCC] dark:fill-[#4F5662]" />
 				</div>
 
-				<div className="absolute opacity-40 right-2">
+				{/* <div className="absolute opacity-40 right-2">
```
evereq, line 228: @YehualashetGit why this got commented out? Not needed anymore?

YehualashetGit: yes we don't need this code s actually,  basically the invite row won't have such menu , may be i will remove it

apps/web/app/constants.ts
YehualashetGit: sure I will rename that

evereq:
@YehualashetGit see my comments and let's fix this quickly so we can merge as soon as possible, thanks!
Also, most probably we will have to make more improvements to the table view later. Honestly the idea of table view for "Issues" was different - it should be that "Issue title/description" would go as the first column, etc, but what you essentially did is exactly the same view we had for "horizontal cards" but now implemented also as a table (more condensed) view and I LIKE that. 

So we will merge this and you will work to make 3rd view for the same data (block view), but essentially we will need to also work on a real "issues" page later. Let's discuss that later in Slack, for now let's finish this PR merge it, and create a "block" view next (from our Figma):

image

apps/web/app/constants.ts
evereq: Btw, I think values should be:
CARDS, TABLE, and BLOCKS, please adjust @YehualashetGit  :)

YehualashetGit: sure

badalkhatri0924:
image

@YehualashetGit there are some UI issues

apps/web/lib/features/team/invite/user-invite-card.tsx
evereq: Yes, makes sense @YehualashetGit

badalkhatri0924:
image

@YehualashetGit UI issue.

apps/web/lib/features/team/invite/user-invite-card.tsx
evereq: @YehualashetGit On the other hand, idea was to make that Invite row to look similar to other rows... Iike it's still "empty" but when you invite someone that person will be shown there with task and other info... So I would prefer you to restore it back :)))

Also in the future when you removing something (even if it looks like not used), please state it in the PR explicitly - we should be very careful with remove things if we not sure about initial idea why it was added etc.

YehualashetGit: okay understood , the main reason why I removed it was , there was no real data to display and they are just a place holder , that's why , but will reverted it back

evereq:
@YehualashetGit you working on fixes for those issues right?

apps/web/lib/features/team/invite/user-invite-card.tsx
YehualashetGit: @evereq  is this a temporary or it only be visible on Table view cause ,it'll now  make me to add more logic on TeamMeberTable View component to the invite row with separate actions

YehualashetGit:
> @YehualashetGit you working on fixes for those issues right?

yes I've fixed the ui issues

evereq:
@badalkhatri0924 can you review again please? If something still broken try to quick fix yourself and let's merge this one so I can review it online

badalkhatri0924 (approved):

evereq:
@YehualashetGit I merged with "squash", so please don't use anymore that same branch, create a new one based on develop for all future work on those tasks. Thanks!

PR #1657, opened 2023-10-26
[DevOps] Deploy to Fly.io 

paradoxe35:
- Fix #1657

Screenshot 2023-10-26 at 13 17 41

### Create a Fly application if it does not exist

```bash
flyctl launch
```

### Deploy
```bash
flyctl deploy
```

### Application crashed because it ran out of memory

If you get the following error: `
Out of memory: Killed process 322 (node) total-vm:569056kB, anon-rss:47632kB, file-rss:4kB, shmem-rss:0kB, UID:0 pgtables:788kB oom_score_adj:0
`

Run the following command (add more RAM):

```bash
flyctl scale memory 1024 -a ever-teams
```


fly.toml
```diff
@@ -0,0 +1,19 @@
...
+#
+# See https://fly.io/docs/reference/configuration/ for information about how to use this file.
+#
+
+app = "ever-teams"
+primary_region = "lax"
+kill_signal = "SIGINT"
+kill_timeout = "5m0s"
+
+[build]
```
paradoxe35, line 11: @evereq 

I think we should use the image link which is already built and pushed (GitHub or Docker Hub)
Just like you did in gauzy

https://github.com/ever-co/ever-gauzy/blob/e875f02e4e9f417252c3322a5cc8311d3372476f/.fly/applications/webapp/fly.toml#L5-L8

evereq: @paradoxe35  yes, so we have 3 ways to make it work for Ever Teams:

1) one is to use Docker and run docker in Fly.io (and other platforms), similarly to how we do it in Ever Gauzy

2) maybe Fly.io support natively running NodeJs apps? If so, it's basically similar to Vercel.com, so we should use that instead of Docker for Fly.io, but we MAY need Docker anyway for other platforms.

3) we can build Ever Teams maybe as "static" ("export" from NextJs) and deploy into some CDN in Fly.io and others (e.g. to Storage like S3 with Cloudfront in front of it, of course, what is available in Fly.io etc). However, I am NOT sure at this stage Ever Teams will work like that, most probably we are using some SSR already or something like that, hard to tell. 
 
So yes, if we do it as 1) above, we probably have to push to some public registry Docker image and run it in Fly.io and other services. You can find relevant GitHub actions inside Gauzy project

apps/web/next.config.js
```diff
@@ -14,9 +24,6 @@ const nextConfig = {
 			'apistage.gauzy.co',
 			'gauzy.s3.wasabisys.com'
 		]
-	},
-	experimental: {
```
evereq, line 18: @paradoxe35 why we remove that in this PR?

Dockerfile
```diff
@@ -0,0 +1,58 @@
+# syntax = docker/dockerfile:1
+
+# Adjust NODE_VERSION as desired
+ARG NODE_VERSION=18.17.1
+FROM node:${NODE_VERSION}-slim as base
+
+LABEL fly_launch_runtime="Next.js"
```
evereq, line 7: @paradoxe35 is that required in Dockerfile? Because this same dockerfile can be used not only in Fly.io but in other providers, so we would prefer to keep the Dockerfile "clean" from anything fly-specific. If it's required, maybe we can specify it some other way in fly config or maybe we need to create 2 Dockerfiles, one for Fly and another "common"?

apps/web/next.config.js
paradoxe35: In NextJS 13 this setting is not available
Since the AppDir is no longer experimental, it is already included in the module core

Also It could raise error while building image

@evereq

Dockerfile
paradoxe35: Don't think it's required
Let me remove it and try to deploy 

@evereq

fly.toml
paradoxe35: From what I have seen so far
I works well with Docker

Dockerfile
paradoxe35: Just removed that line `LABEL fly_launch_runtime="Next.js"`

fly.toml
paradoxe35: I'm going to create GitHub actions then @evereq

evereq: Ok, I will add required secrets on Github, let me know list of env var names

paradoxe35: Sure @evereq

evereq:
@paradoxe35 I merged this with "squash", please create another branch based on "develop" for any additional things, do NOT reuse this branch. Thanks

PR #1671, opened 2023-10-27
feat:add render yaml file

maceteligolden:


maceteligolden:
@evereq please what do you think about using something like this for github actions for render https://github.com/marketplace/actions/render-deploy-action

evereq:
> @evereq please what do you think about using something like this for github actions for render https://github.com/marketplace/actions/render-deploy-action

I will check it more, yes saw that action before... For now merging your PR and will test etc

PR #1754, opened 2023-11-09
Fix/updated GitHub integration UI

badalkhatri0924:
image


evereq:
Hm, but what about text of description for each? Is it "Lorem" like in screenshot!? :)

evereq:
Also, text is "hardcoded" in the components code...

evereq:
@badalkhatri0924 see also https://github.com/ever-co/ever-teams/actions/runs/6815310943/job/18534123124#step:7:50

badalkhatri0924:
@evereq fixed all.

PR #1771, opened 2023-11-13
Update packages sentry

NdekoCode:
Updated sentry configuration on next.js
- Update @sentry/nextjs to 7.80.0
- Added sentry debug page
- Added sentry debug route
- Custom source map file with webpack compiler and Sentry AUTH_TOKEN


apps/web/pages/api/sentry-example-api/index.ts
```diff
@@ -0,0 +1,9 @@
+import { NextApiRequest, NextApiResponse } from 'next';
+
+export default function handler(req: NextApiRequest, res: NextApiResponse) {
+	try {
+		throw new Error('Sentry Example API Route Error');
```
evereq, line 5: @NdekoCode why do we need this code / example page??

apps/web/sentry.client.config.ts
```diff
@@ -28,3 +29,7 @@ Sentry.init({
 		})
 	]
 });
+// if (process.env.NODE_ENV === 'production' && process.env.SENTRY_DSN) {
```
evereq, line 32: @NdekoCode why this code commented out!? Also I think this code using @sentry/node page which is for SERVER, but the file called sentry.client! So it's in wrong place and commented out!

apps/web/next.config.js
```diff
@@ -25,44 +26,68 @@ const nextConfig = {
 			'apistage.gauzy.co',
 			'gauzy.s3.wasabisys.com'
 		]
+	}, // Optional build-time configuration options
+	sentry: {
+		// See the sections below for information on the following options:
```
evereq, line 31: @NdekoCode very weird that tons of those lines are commented out, but we have same lines below in this file! E.g. tunnelRoute in https://github.com/ever-co/ever-teams/pull/1771/files#diff-a5de0fa0d80f14a44e9bdf2ed5d0cbb3a7bd5fdcacfbc1d03b5dcb57fc4943c1R86

apps/web/.env
```diff
@@ -47,6 +47,8 @@ MEET_JWT_APP_SECRET=
 # SENTRY
 SENTRY_ORG=ever-co
 SENTRY_PROJECT=ever-teams-web
+SENTRY_LOG_LEVEL=[info|debug]
```
evereq, line 50: @NdekoCode is it possible to specific both values that way?? Or it should be one of them!?

NdekoCode: It should be one of them, Mr. @evereq, it's recommended during the build so that sentry can build the final configuration.

apps/web/next.config.js
NdekoCode: Understanding the sentry configuration was difficult for me, there are comments that come with the initial configuration but there are others that I added at the time of testing.
I have these comments to help me and those who will want to understand the updates, there are also several configuration concerning the build of the application like the options:
// disableServerWebpackPlugin: true,
// disableClientWebpackPlugin: true,
  which you need to uncomment if you encounter a permission problem during build, for example, if you don't have a SENTRY_AUTH_TOKEN.

apps/web/pages/api/sentry-example-api/index.ts
NdekoCode: It may happen that the sentry configuration doesn't work, so this page is generated to test whether we receive errors in the sentry dashboard. We can delete it once you've checked that the configuration on your side is working properly, sir.

apps/web/sentry.client.config.ts
NdekoCode: This code was there in the test version, when I wanted to check that you can only run sentry in production mode, but I remembered that DNS will only be provided when you test or when you put the application into production.
but I left it there in case you might decide otherwise, sir @evereq

apps/web/sentry.server.config.ts
```diff
@@ -5,12 +5,15 @@
...
-Sentry.init({
-	dsn: SENTRY_DNS,
 
-	// Adjust this value in production, or use tracesSampler for greater control
-	tracesSampleRate: 1,
-
-	// Setting this option to true will print useful information to the console while you're setting up Sentry.
-	debug: process.env.NEXT_PUBLIC_SENTRY_DEBUG && process.env.NEXT_PUBLIC_SENTRY_DEBUG === 'true' ? true : false
-});
+if (SENTRY_DNS) {
```
NdekoCode, line 9: Sir @evereq SENTRY_DSN is not defined but rather SENTRY_DNS

NdekoCode (commented):
I'm going to implement the same logic in the other sites.

PR #1804, opened 2023-11-16
Fix/api calls

Sergemuhundu:


apps/mobile/ios/GauzyTeams/Info.plist
```diff
@@ -30,7 +30,7 @@
 		</dict>
 	</array>
 	<key>CFBundleVersion</key>
-	<string>7</string>
+	<string>8</string>
```
evereq, line 33: @Sergemuhundu you forgot to increase this one :) Also, why we have still folder name as "GauzyTeams", not "EverTeams"? Can we rename ALL that in next PR please?

PR #1821, opened 2023-11-17
fix spelling

NdekoCode:
# fix spelling

Added also the `cspell` package so that other team members can check for spelling errors during development with the `yarn spellcheck` command.


evereq:
@NdekoCode pre-build failing because you did not pushed updated yarn.lock file?

NdekoCode:
I'm in to see sir @evereq , but when I checked the staging area of git, this file had not been modified.

PR #1836, opened 2023-11-18
Bug/expo doctor issues fix

evereq:


evereq:
@desperado1802 I know you did not finished yet, but I want to push it to see how it works anyway ;)

PR #1846, opened 2023-11-21
Bug/expo doctor fixes

### LINKED ISSUE #1825: [BUG] Fix Expo Doctor Issues/Warnings

desperado1802:
#1825
Resolved most of the warnings of expo doctor
image


desperado1802:
The first warning from 2 is regarding  @expo/config-plugins. Already updated it, but for some reason in yarn.lock we still have both old and new version. If we delete and install yarn.lock it should get rid of the old version, however there will be a lot of package updates. Should we do it @evereq ?

evereq:
> The first warning from 2 is regarding @expo/config-plugins. Already updated it, but for some reason in yarn.lock we still have both old and new version. If we delete and install yarn.lock it should get rid of the old version, however there will be a lot of package updates. Should we do it @evereq ?

Should we do it @evereq ? Yes, let's do in another PR and let's use another branch for that, don't reuse this branch anymore. Thanks!

PR #1847, opened 2023-11-21
Call Backend API based of optional NEXT_PUBLIC_GAUZY_API_SERVER_URL Variable

badalkhatri0924:
https://github.com/ever-co/ever-teams/assets/81486442/93fff9de-a823-4ca2-9cf3-91c743408557


apps/web/app/services/client/api/task-sizes.ts
```diff
@@ -21,6 +22,23 @@ export function deleteTaskSizesAPI(id: string) {
 	return api.delete<DeleteResponse>(`/task-sizes/${id}`);
 }
 
-export function getTaskSizesList(tenantId: string, organizationId: string, activeTeamId: string | null) {
-	return api.get(`/task-sizes?tenantId=${tenantId}&organizationId=${organizationId}&activeTeamId=${activeTeamId}`);
+export async function getTaskSizesList(tenantId: string, organizationId: string, activeTeamId: string | null) {
+	const endpoint = `/task-sizes?tenantId=${tenantId}&organizationId=${organizationId}&organizationTeamId=${activeTeamId}`;
+
+	// API call direct to Backend Server
+	if (process.env.NEXT_PUBLIC_GAUZY_API_SERVER_URL) {
```
evereq, line 29: @badalkhatri0924 can we put all that code `if (process.env...)...` inside of `api.get(...)` method? So that we don't duplicate SAME in ALL methods like that? We can also add one more parameter "isDirect" to the api.get signature like this:

```
return api.get(endpoint, true) 
```

So that methods that should work with direct way will do it and other methods where we still have `return api.get(endpoint) ` will not need to do it? etc.

Basically my concern is that all methods like `getTaskSizesList` will not have lots of additional not relevant code, while we can use either "factory" pattern or just "abstract" direct / not direct inside api.get method itself... Btw, same for all other methods etc.

badalkhatri0924: @evereq please review this again.

apps/web/app/services/client/api/task-sizes.ts
```diff
@@ -21,6 +21,9 @@ export function deleteTaskSizesAPI(id: string) {
 	return api.delete<DeleteResponse>(`/task-sizes/${id}`);
 }
 
-export function getTaskSizesList(tenantId: string, organizationId: string, activeTeamId: string | null) {
-	return api.get(`/task-sizes?tenantId=${tenantId}&organizationId=${organizationId}&activeTeamId=${activeTeamId}`);
+export async function getTaskSizesList(tenantId: string, organizationId: string, activeTeamId: string | null) {
+	const endpoint = `/task-sizes?tenantId=${tenantId}&organizationId=${organizationId}&organizationTeamId=${activeTeamId}`;
+
+	// API call via Proxy Nextjs /api routes
+	return process.env.NEXT_PUBLIC_GAUZY_API_SERVER_URL ? apiDirect.get(endpoint) : api.get(endpoint);
```
evereq, line 28: @badalkhatri0924 still, I don't get why we can't encapsulate that check for `process.env.NEXT_PUBLIC_GAUZY_API_SERVER_URL` inside some api.get() method?? 

get(endpoint, isDirect) { 
  return isDirect && process.env.NEXT_PUBLIC_GAUZY_API_SERVER_URL ? apiDirect.get(endpoint) : api.get(endpoint);
}

and so in the method above just do return api.get(endpoint, true); when we know that method works with direct and do `api.get(endpoint)` as before when we know when method don't work yet with direct???

badalkhatri0924: @evereq please check again, I've updated code.

evereq:
@badalkhatri0924 The last variant looks good, I merged with Squash, so all the next work will be done in another branch and delegated to another developer. Thank you!

PR #1875, opened 2023-11-24
Feat/add kanban darkmode

maceteligolden:
Screen Shot 2023-11-24 at 7 47 56 PM


evereq:
@maceteligolden looks great. Can we make sure that this page is shown together with the site header and other details? I.e. not on empty page :))

PR #1949, opened 2023-12-03
Feat/task block view for issues

Cedric921:
- display issues in block view
Screenshot from 2023-12-03 19-34-17
Screenshot from 2023-12-03 19-34-23
Screenshot from 2023-12-03 19-34-31


apps/web/lib/features/team-members.tsx
```diff
@@ -68,6 +76,17 @@ export function TeamMembers({ publicTeam = false, kanbanView: kanbanView = Issue
 				</Transition>
 			);
 			break;
+
+		case kanbanView == IssuesView.BLOCKS:
```
evereq, line 80: @Cedric921 why `kanbanView`, should be probably `blocksView`, no?

apps/web/package.json
```diff
@@ -22,6 +22,7 @@
 		"@heroicons/react": "^2.0.12",
 		"@jitsi/react-sdk": "^1.3.0",
 		"@jitsu/jitsu-react": "^1.3.0",
+		"@nextui-org/react": "^2.2.9",
```
evereq, line 25: @Cedric921 I love nextui lib, but I wonder how much more effort will be needed to reuse existed libs we already reference (e.g. https://github.com/shadcn-ui/ui, etc)?

PR #2052, opened 2024-01-03
fix: display language list in dropdown on personnal settings page

Cedric921:
Screenshot from 2024-01-03 16-33-53


apps/web/lib/settings/personal-setting-form.tsx
```diff
@@ -137,6 +137,8 @@ export const PersonalSettingForm = () => {
 					id: user.id
 				});
 			}
+
+			window.location.replace(`/${newLanguage}/settings/personal`);
```
evereq, line 141: @Cedric921 is this the right way to route to another "page"? I though you need some router or something...

Cedric921: I change it to next router

PR #2065, opened 2024-01-05
[Refact] move API routes from `page/api` to `app/api` folder

Cedric921:


evereq:
@Cedric921 also make sure you fixing periodically those "DeepScan" issues as you go :)

PR #2070, opened 2024-01-06
Refact: change Texts and Titles on Auth pages

Cedric921:
- We have change some textes in auth pages and fixed columns header of Team Member cards
Screenshot from 2024-01-06 23-17-40


evereq:
@Cedric921 "Open Work and Project Management Platform" should fit into one line :)

evereq:
@Cedric921 Also, we don't need an "arrow" near the text "Register" unless there is where to go "back" from this page, can you check ? Mean, is there any logic related to that back arrow like if there is referral maybe or maybe some query string and user click on that arrow it goes BACK to previous page etc?

Cedric921:
> @Cedric921 Also, we don't need an "arrow" near the text "Register" unless there is where to go "back" from this page, can you check ? Mean, is there any logic related to that back arrow like if there is referral maybe or maybe some query string and user click on that arrow it goes BACK to previous page etc?

fixed

apps/web/app/[locale]/auth/passcode/page.tsx
```diff
@@ -101,10 +101,13 @@ function EmailScreen({ form, className }: { form: TAuthenticationPasscode } & IC
 
 					<div className="flex items-center justify-between w-full mt-6">
 						{/* Send code */}
-						<div className="flex flex-col items-start">
-							<Link href="/auth/team">
-								<BackButton />
-							</Link>
+						<div className="flex flex-col items-start gap-2">
+							<div className="flex justify-start items-center text-sm">
+								<span>{t('common.BACK')}</span>
```
Comment (evereq) [inline apps/web/app/[locale]/auth/passcode/page.tsx:106]: @Cedric921 can you pay attention to small details? E.g. common.BACK is NOT "Back" anymore, it's `common.DONT_HAVE_ACCOUNT` or something like that...

apps/web/app/[locale]/auth/passcode/page.tsx
```diff
@@ -101,10 +101,13 @@ function EmailScreen({ form, className }: { form: TAuthenticationPasscode } & IC
...
+								<Link href="/auth/team" className="text-primary">
+									<span>Register now</span>
```
Comment (evereq) [inline apps/web/app/[locale]/auth/passcode/page.tsx:108]: Why "Register now" is not in the translation files? Also, again small detail - it should be "Register Now!".

apps/web/lib/components/button.tsx
```diff
@@ -92,12 +91,13 @@ export function RoundedButton({ children, className, ...rest }: RoundedButtonPro
...
 export function BackButton({ onClick, className }: { onClick?: () => void; className?: string }) {
-	const t = useTranslations();
 
 	return (
-		<button type="button" className={clsxm('flex items-center', className)} onClick={onClick}>
+		<button type="button" className={clsxm('flex items-center justify-start text-sm', className)}>
 			<ArrowLeft className="mr-2" />
-			<span className="text-sm">{t('common.BACK')}</span>
+			<span className="text-sm" onClick={onClick}>
+				Back
```
evereq, line 99: Here @Cedric921 you actually should use common.BACK with value "Back" etc.

apps/web/lib/i18n/en.ts
```diff
@@ -167,11 +167,11 @@ export const en = {
 			INPUT_INVITE_CODE: 'Input invitation code.',
 			INPUT_INVITE_CODE_DESC: 'Enter the invitation code we sent to your email.',
 			INVALID_INVITE_CODE_MESSAGE: 'Invalid Code',
-			WELCOME_TEAMS: 'Welcome to Ever Teams',
+			WELCOME_TEAMS: 'Real-Time Clarity, Real-Time Reality™.',
 
-			COVER_TITLE: 'Follow your teams work progress in real-time!',
+			COVER_TITLE: 'Open Work  and Project Management Platform',
```
evereq, line 172: Why did you put 2 spaces here before "and"?

Cedric921: Fixed

apps/web/lib/components/button.tsx
Cedric921: resolved

PR #2082, opened 2024-01-11
Feat: display details of time slots & create API route to delete time slots

Cedric921:
- Screenshots page Skeletons
Screenshot from 2024-01-11 22-21-36

- Detail Slot Pop-up (Light)
Screenshot from 2024-01-11 22-24-33

- Detail Slot Pop-up (Dark)
Screenshot from 2024-01-12 00-06-44


.nx/cache/nx-console-project-graph/project-graph.html
```diff
@@ -0,0 +1,33 @@
+<!DOCTYPE html>
```
evereq, line 1: @Cedric921 do we need that nx cache folder in repo?

PR #2109, opened 2024-01-22
Feat: activity for Task

Cedric921:
-  Connected User
Screenshot from 2024-01-23 23-17-29

- Other User
Screenshot from 2024-01-23 23-40-20

- Other Activity
Screenshot from 2024-01-23 23-18-02

- Error page
errer


evereq:
@Cedric921 I think somehow when I merged "maintanance" PR, it was including a lot of this PR etc.
Can we somehow merge all of that as soon as possible so we don't have merge issues later etc? 
Preferable today, even if something not finished, but not broke anything etc

PR #2137, opened 2024-01-30
fix: errors

Cedric921:
- fix fonts ( By using next/font/google

### DIFF
--- a/apps/web/next.config.js
+++ b/apps/web/next.config.js
@@ -55,6 +55,19 @@ const nextConfig = {
 			'gauzystage.s3.wasabisys.com'
 		]
 	},
+	// module: {
+	// 	rules: [
+	// 	{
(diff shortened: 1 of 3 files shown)


evereq:
@Cedric921 how is it going with this PR?

apps/web/next.config.js
```diff
@@ -55,6 +55,19 @@ const nextConfig = {
 			'gauzystage.s3.wasabisys.com'
 		]
 	},
+	// module: {
```
evereq, line 58: @Cedric921 it's not needed? Why we keep that commented code in such case?

evereq:
@Cedric921 hm, I think this PR is not rebased well as it contains many other changes / commits etc?

Cedric921:
> @Cedric921 hm, I think this PR is not rebased well as it contains many other changes / commits etc?

Now Fixed

PR #2171, opened 2024-02-06
2165 improvement add dotted types of translation

Anishali2:
Added the translations dotted types


evereq:
@Anishali2 it tells 14 new issues in DeepScan, can you fix those?

Anishali2:
It is not merged by develop branch so that's why he is saying those issues

Anishali2:
those issues are already fixed in develop branch

PR #2190, opened 2024-02-09
fix(Localization): fix the localization of web app

Anishali2:
Sir i've created a new file in web/types/local.ts which is not doing anything for the project but it through some error if some translation exist in `en` and not exist in other langauges


Anishali2:
Next time if someone wants to add translation he should have to add into all the locals so there is default check otherwise typescript throughs some errors

Anishali2:
Every translation have same properties no more no less

PR #2192, opened 2024-02-09
Feat: Data Synchronization Mode UI on Setting Page

Cedric921:
- When `Auto Sync` is Disabled (As default is Enabled)

Screenshot from 2024-02-09 11-00-48 (Sync Desabled)

- When Use Enable `Auto Sync`

Screenshot from 2024-02-09 11-00-42( Sync enabled)

- Popup When User Try to Select `Real Time`

Screenshot from 2024-02-09 15-41-19(Real TimePopu)


evereq:
@Cedric921 just want to be sure - by default Auto Sync should be enabled :)

PR #2194, opened 2024-02-10
fix(Kanban): fix the kanban scrolling issue

Anishali2:
https://github.com/ever-co/ever-teams/assets/105014311/a9f5f8ca-6073-4c3d-b463-46527cfa97a0


evereq:
@Anishali2 please check that other pages (e.g. issue page or member profile page) are still work well (I did not tested, but I think might be some issues etc).

evereq:
@Anishali2 also test both light / dark theme etc

Anishali2:
Sir there is an issue in actions webapp but when i try to open webapp it throughs exception error and I'm unable to figureout this errors

Anishali2:
> @Anishali2 also test both light / dark theme etc

Sir i've tested the light and dark mode so it wont be an issue here

Anishali2:
This is user profile page I don't see any issues here should i have to move the header tabs here ?

image

evereq:
> This is user profile page I don't see any issues here should i have to move the header tabs here ?
> 
> image

No, no need to move header tabs to that page and to Task (issue) page too etc. Only needed those switches on main page

PR #2209, opened 2024-02-13
2208 fix kanban scrolling

Anishali2:
[full-scroll-kanban.webm](https://github.com/ever-co/ever-teams/assets/105014311/847bfa0c-9092-4510-bba9-a9452c11fe31)

### DIFF
--- a/apps/web/lib/features/team-members-kanban-view.tsx
+++ b/apps/web/lib/features/team-members-kanban-view.tsx
@@ -1,7 +1,6 @@
 import { useKanban } from '@app/hooks/features/useKanban';
 import { ITaskStatus, ITaskStatusItemList, ITeamTask } from '@app/interfaces';
 import { IKanban } from '@app/interfaces/IKanban';
-import { fullWidthState } from '@app/stores/fullWidth';
 import { clsxm } from '@app/utils';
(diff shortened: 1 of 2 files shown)


apps/web/lib/features/team-members-kanban-view.tsx
```diff
@@ -235,20 +230,12 @@ export const KanbanView = ({ kanbanBoardTasks }: { kanbanBoardTasks: IKanban })
 									) : null}
 									<>{provided.placeholder}</>
 								</div>
-								<Container fullWidth={fullWidth} className={clsxm('w-full !mx-0 px-8')}>
```
evereq, line 238: @Anishali2 where this "footer" go now? I can't see it in PR, but I see it visible in the video you recorded, so how we display it if you removed that code here?

Anishali2: It's in the main-layout I recently removed added the class footer `footerClassName` and from main layout of kanban screen and so it's visible in the main app screen so we don't need it anymore

Anishali2: I've done this before for scrolling and now the requirements are adding scroll on the full screen so we actually don't need it anymore

PR #2215, opened 2024-02-15
[Fix]: invalid code error fix

Anishali2:
This error is not working because api is returning response `code:201` with message `"Unauthorized"` 
key: https://api.ever.team/api/auth/signin.email/confirm?includeTeams=true
So I just check that if the message is unauthorized or not 
image


Anishali2:
[login-invalid-code.webm](https://github.com/ever-co/ever-teams/assets/105014311/0813ed24-3f14-494c-adcf-e8a5110ef6c4)

PR #2220, opened 2024-02-17
[UI]: design card of kanban is fixed

Anishali2:
before
image

after
image


Anishali2:
UI improvement list

- time progress bar hiding fix
- title inline added
- collabraters images compressing 
- tags added 
- dynamic hights of boxes

PR #2226, opened 2024-02-19
[UI]: Settings page design re-arrange and added the layout system

Anishali2:


Anishali2:
Before

[settings-screen-before-view.webm](https://github.com/ever-co/ever-teams/assets/105014311/2e0dd3d5-0b29-48d3-9ac3-303d901a1e04)






After

[settings-screen-view.webm](https://github.com/ever-co/ever-teams/assets/105014311/d7ce7952-63e9-4d39-8229-e3e7a3b14c69)

Anishali2:
I've setup the layout system in setting page so that we can we can optimize the performance of our app and used one component for both pages

PR #2261, opened 2024-03-03
[Fix]: working on kanban section functionality

Anishali2:


Anishali2:
Sending the video of which changes are implemented for better understanding.

apps/web/app/[locale]/kanban/page.tsx
```diff
@@ -47,6 +50,10 @@ const Kanban = () => {
 		{ name: t('common.YESTERDAY'), value: KanbanTabs.YESTERDAY },
 		{ name: t('common.TOMORROW'), value: KanbanTabs.TOMORROW }
 	];
+	const { user } = useAuthenticateUser();
+	const { openModal, isOpen, closeModal } = useModal();
+	const timezone = userTimezone();
+	console.log('time-zone', timezone);
```
Comment (evereq) [inline apps/web/app/[locale]/kanban/page.tsx:56]: @Anishali2 console.log?

apps/web/lib/components/kanban-card.tsx
```diff
@@ -216,32 +241,31 @@ export default function Item(props: ItemProps) {
...
-				<div className="w-full flex items-center justify-between">
-					<div className="mt-1">
+				<div className="w-full h-10 flex items-center justify-between">
+					<div>
 						{item.status === TaskStatus.INPROGRESS ? (
 							<div className="flex items-center gap-2">
 								<small className="text-grey text-xs text-normal">Live:</small>
 								<p className="text-[#219653] font-medium text-sm">
-									{pad(hours)}:{pad(minutes)}:{pad(seconds)}{' '}
+									{h}h: {m}:m
```
evereq, line 256: @Anishali2 I think here idea was to show "seconds" and with continues "refresh" it will look like "real-time" data etc. Even if we did not use yet such real-time updates, I would like to keep that idea. But correct me if I am wrong and this place makes sense only show minutes/hours, not seconds?

Anishali2: Sir previously it shows gauzy-app-timer time which can render whole kanban after every milliseconds but now I've shown task total worked time..

Anishali2: I've shown the seconds also

Anishali2:
Add Member
User Stack List Imporved
Remaining users list 
Timer Countdown

[kanban-page.webm](https://github.com/ever-co/ever-teams/assets/105014311/aa7c4fa2-8b72-4148-ba11-b1c04a938024)

apps/web/app/[locale]/kanban/page.tsx
Comment (Anishali2) [reply apps/web/app/[locale]/kanban/page.tsx:56]: removed..

Comment (Anishali2) [reply apps/web/app/[locale]/kanban/page.tsx:56]: Sir the is configuration in linting in which user have to remove console every time before pushing code but the reason is if I add this configuration I have to remove every single console in the project in which some consoles are useful

Anishali2:
All checks are resolved 
Sir please merge this PR so I can further work on this page...

apps/web/app/[locale]/kanban/page.tsx
Comment (evereq) [reply apps/web/app/[locale]/kanban/page.tsx]: Yes @Anishali2 , some are useful 100%, maybe we can add some "hint" to linter above those so it ignores them?

apps/web/lib/components/kanban-card.tsx
evereq: @Anishali2 I don't get that part "previously it shows gauzy-app-timer time" but " now I've shown task total worked time", you mean you decide to show one value instead of another? Forget for a second about issue with "refresh" etc, I am asking logically you decided to show some DIFFERENT time? Explain please exactly what time was shown before and what time will be shown now from point of view of manager who overview that board data.

apps/web/app/[locale]/kanban/page.tsx
```diff
@@ -67,7 +73,10 @@ const Kanban = () => {
 							{t('common.KANBAN')} {t('common.BOARD')}
 						</h1>
 						<div className="flex w-fit items-center space-x-2">
-							<strong className="text-gray-400">08:00 ( UTC +04:30 )</strong>
+							<strong className="text-gray-400">
+								{`(`}
+								{timezone.split('(')[1]}
```
Comment (evereq) [inline apps/web/app/[locale]/kanban/page.tsx:78]: @Anishali2 I don't understand what that suppose to show? Can you talk with @KostyaEver to ask him what was idea of that time / timezone etc? What is that value you display and why it's interesting for user to see?

PR #2267, opened 2024-03-04
Fix: minors

Cedric921:


apps/web/lib/features/team/user-team-table/user-team-table-header.tsx
```diff
@@ -3,11 +3,11 @@ import React from 'react';
 function UserTeamTableHeader() {
 	return (
 		<div className="flex overflow-y-auto px-4 text-center items-center pt-4">
-			<p className="min-w-[8rem]  w-1/3 ">Name</p>
-			<p className="min-w-[8rem] w-2/5 ">Task</p>
-			<p className="min-w-[8rem] w-1/12 ">Worked on Task</p>
-			<p className="min-w-[8rem] w-1/4 ">Estimate</p>
-			<p className="min-w-[8rem] w-28 ">Action</p>
+			<p className="min-w-[8rem] 2xl:w-[20.625rem] w-1/4">Name</p>
```
evereq, line 6: @Anishali2 those titles like "Task", "Worked on Task" etc should also come from translation files.
CC: @Cedric921 in the future watch out for such things too, after @Anishali2 will move all existed

PR #2306, opened 2024-03-18
[Fix]: epic and labels filter added

Anishali2:


apps/web/lib/features/task/task-status.tsx
Anishali2, line 1: this ',' is added in my prevoius PR where I modified the icon structure of the project so I just removed that

Anishali2: this ',' added by mistake

PR #2342, opened 2024-03-29
[Design]: design setting and re-fetch

Anishali2:
[setting-view.webm](https://github.com/ever-co/ever-teams/assets/105014311/af13c54b-c108-4be2-83a7-a704725d2f3c)

### DIFF
--- a/apps/web/app/[locale]/settings/layout.tsx
+++ b/apps/web/app/[locale]/settings/layout.tsx
@@ -11,6 +11,7 @@ import Link from 'next/link';
 import React from 'react';
 import { useRecoilState, useRecoilValue } from 'recoil';
 import { clsxm } from '@app/utils';
+import { withAuthentication } from 'lib/app/authenticator';
 
 const SettingsLayout = ({ children }: { children: JSX.Element }) => {
 	const t = useTranslations();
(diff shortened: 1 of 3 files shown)


evereq:
@Anishali2 I remember we used very thin scrollbar long time ago, not sure what happened with that, any ideas?
It was something like this https://adoxography.github.io/tailwind-scrollbar/examples#global-scrollbar-colours (thin one) or something like in Gauzy:

image

Because scrollbar you have for left sidebar looks VERY bad...

Anishali2:
This is the one from shadcn that provide small scroll bar.
https://ui.shadcn.com/docs/components/scroll-area

Anishali2:
image

evereq:
Yes, so either one from shadcn or from https://github.com/adoxography/tailwind-scrollbar will work for us I think...

Anishali2:
Can I add this into settings page or in future replace all the scrollbars of project into these scrollbars after kanban board

evereq:
> Can I add this into settings page or in future replace all the scrollbars of project into these scrollbars after kanban board

Not sure, but yes, I think we better to use such scrollbars everywhere where it make sense!
I.e. maybe to scroll main content it's fine to use standard, but if page has more than 1 scrollbar horizontal or more than 1 scrollbar vertical, I would prefer to use at least one of them to be thin one etc.

Anishali2:
> Not sure, but yes, I think we better to use such scrollbars everywhere where it make sense! I.e. maybe to scroll main content it's fine to use standard, but if page has more than 1 scrollbar horizontal or more than 1 scrollbar vertical, I would prefer to use at least one of them to be thin one etc.

We mostly have 2 scroll bars across all pages

evereq:
@Anishali2 great, please create another PR that adds thin scrollbars in right places. Preferable first to try to add thin for blocks that are not main content, e.g. in the settings page, we probably want thin scrollbar inside left sidebar, but we don't want for now to change main scrollbar of the browser (located at right side of screen)

Anishali2:
Ok Sir

PR #2351, opened 2024-04-01
2350 fix linting and warnings

Anishali2:


apps/web/env.js
```diff
@@ -1,30 +1,30 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
 require('dotenv').config();
 
-const port = process.env.PORT || 3030;
```
evereq, line 4: @Anishali2 why you commented this whole file code out? Do you know why that file was needed etc?

evereq: No, please uncomment it now and this logs are needed because they inform user on which port it's running etc!

Anishali2: Ok Sir give me a sec

PR #2361, opened 2024-04-05
[Fix]: task status ids are fixed in kanban board

Anishali2:


Anishali2:
Sir, now you can create a new task on the home page it will show that task in Kanban but it only works with newly created tasks

evereq:
@Anishali2 How can we "fix" existed tasks? Maybe @rahul-rocket can help to create some data migration in Gauzy that goes over all tasks and "fix" they status ID according to status name, is that what we need to see tasks in Kanban?

Anishali2:
Sir, there is only the status name is stored and if the name is not changed then it is possible if the name is changed then it's hard for anyone to figure out which status it is saved before because now there is no status exist

rahul-rocket:
@Anishali2 @evereq Okay. Atleast we can set status ID for those statuses which is not changed yet. 

Note: We need to create migration for this.

evereq:
@rahul-rocket yes, so you mean like iterate over all tasks that exists, see if "statuses" table have same name (for the same tenant / org) as saved in "task" and if yes, set statusId in the task to correct value? You mean migration like that?

PR #2379, opened 2024-04-18
[Bug]: Re-rendering issue in user profile page

Anishali2:


apps/web/app/hooks/useInfinityFetch.ts
Anishali2, line 1: this useInfinityFetch hook useEffect calls when the user scrolls to the bottom but if no items exist then it is already to the bottom so it calls the hook infinite time which makes the application slow because of the re-rendering.
So I just added the if check if the existing data is greater than one then this hook should be working and I also checked that the re-rendering is causing issues because of this problem.

PR #2385, opened 2024-04-22
kanban: remove the list and close modal on save

Anishali2:
List of tasks in kanban create task modals is hide
Auto close the modal when the new task is created


evereq:
@Anishali2 please see https://capture.dropbox.com/MWhQ5TGH473RN9qE and let me know which popup you changed, for "create new task" or for "assign task"? Preferable to always add a short video so we can understand what you did.

Anishali2:
The create task modal is only used in Kanban board so it won't affect any thing 

Give me some time
 - I'll expand the scroll bar in the assigned tasks model.
 - Also the description and dropdowns will be visible by default.

PR #2414, opened 2024-04-25
fix(Task-Issue): fix the issue of task not saving

Anishali2:


apps/web/lib/features/task/task-input-kanban.tsx
```diff
@@ -152,16 +161,17 @@ export function TaskInputKanban(props: Props) {
 			datas.hasCreateForm &&
 			datas
 				.handleTaskCreation({
-					autoActiveTask,
+					autoActiveTask: true,
```
evereq, line 164: @Anishali2 seriously!? So the code below (line 147) someone added because has nothing to do?

```
/*
		If task is passed then we don't want to set the active task for the authenticated user.
		after task creation
	 */
	const autoActiveTask = props.task !== undefined ? false : true;
```

Anishali2: Sir, I've added this before.

I've copied this code from the home page to create task functionality which is written by @paradoxe35 and those functionalities have multiple additional features that are only needed on the home page. But I only want to create tasks in the Kanban board I just filter out some of the things

evereq: Yeah, a better approach would be to move into some component that code and reuse on that original page and in Kanban instead of copying code over :) But I got it, thanks for answer, for now it's OK :)

Anishali2: I verified this twice before changing and created tasks on the home page to verify what is happening and also on the kanban board

Anishali2: The home page creates task by default sets the previous set value but Kanban needs to create a new one always that's why I also removed the value field in Kanban Input at ( line no 239 previously )

PR #2436, opened 2024-04-27
[FIX] fix misaligned timezone field where the dropdown had excessive …

GedeonTS:
Before:

Screenshot from 2024-04-24 16-55-50

Now:
Screenshot from 2024-04-27 21-29-12


evereq:
@GedeonTS no idea how, but seems you used some other branch as a base!? (e.g. "master"). Look into history above

GedeonTS:
Thanks @evereq , I just noticed that I used main as base instead of develop

PR #2495, opened 2024-05-05
[Improve] Add translations to the team block headers

GedeonTS:
[Screencast from 05-05-2024 22:23:58.webm](https://github.com/ever-co/ever-teams/assets/97834160/03b72edd-1c7f-4107-a94a-107e4ffec810)

Screenshot from 2024-05-05 20-47-04


evereq:
Yes, it seems somehow we end up with 2 different packages for i18n:
1) https://github.com/amannn/next-intl
2) https://github.com/i18next/react-i18next (which itself based on https://github.com/i18next/i18next)

evereq:
It feels like we added https://github.com/amannn/next-intl because it had "app" router support, while https://github.com/i18next/react-i18next only worked with "pages" router at that time... However, now it seems both libs support well app router and even server components etc. So we will have to make a choice and keep only one of the libs :))
I'll merge this PR and create separate issue to address this later!

PR #2518, opened 2024-05-13
Member profile page hide some tab options

GloireMutaliko21:
On settings page : 
- [x] If the connected user has a manager role, then he can see a toggle that allows to share some profile views.

On a user's profile page : 
- [x] If the team's settings allows users to see all tabs, then everyone can see everything 
- [x] On Daily Plan tab, even if all the views are public, only managers and own user profile can update/delete their plans. Others can only see them

### DIFF
--- a/apps/mobile/app/services/interfaces/IOrganizationTeam.ts
+++ b/apps/mobile/app/services/interfaces/IOrganizationTeam.ts
@@ -11,6 +11,7 @@ export interface IOrganizationTeamCreate {
 	tags?: any[];
 	organizationId: string;
 	tenantId: string;
+	shareProfileView?: boolean;
 	public?: boolean;
 	imageId?: string | null;
 	image?: IImageAssets | null;
@@ -30,6 +31,7 @@ export interface IOrganizationTeam {
 	members: any[];
 	tags: any[];
 	id: string;
+	shareProfileView?: boolean;
 	createdAt: string;
 	updatedAt: string;
 	imageId?: string | null;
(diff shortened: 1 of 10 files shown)


evereq:
Hm, interesting,  but I think it's should be a Team Setting if team allows to see those tabs for everyone in the team or only for managers (and of course everyone always can see OWN profile tabs). For example, consider very transparent team like our :) We would like to make sure everyone can go to your profile and see what tasks you worked on today and what you plan to work next, we don't want to hide it. However in some orgs, they may want to hide those details etc. 

Basically I think we can have setting like: "Share Member Worked Tasks and Plans with Team" with "On/Off" switch. 
If "On", we show those tabs for "everyone" in the team, if "off", we show only to owner and to managers.
Also double check what @KostyaEver  think about this please :)
Makes sense @GloireMutaliko21 ?

GloireMutaliko21:
You're right, sir @evereq ! I think we'll have to improve the Gauzy API to do this. However, by default, I think it would also be possible to display them for plans, but only limit CRUD options to owners and on managers. 

This could then be configured in the team parameters if the views need to be hidden as well.

evereq:
Great, so @GloireMutaliko21 please work on that APIs and UI in Ever Teams please (I think you just need to add a field to the Team entity right? So we store that boolean setting per team? Or you think more improvements needed!?)

evereq:
@GloireMutaliko21 merged with Squash, make sure you base new PRs on develop branch and not use this branch again

PR #2527, opened 2024-05-17
[Improvement] Remember last issue type while creating a task

JoeMakuta:


apps/web/lib/features/task/task-input.tsx
```diff
@@ -280,6 +280,8 @@ export function TaskInput(props: Props) {
 		}
 	}, [props.autoFocus, targetEl]);
 
+	// const savedIssueType : string | null = localStorage.getItem('savedIssueType') as string && null;
```
evereq, line 283: @JoeMakuta I think it was better to define vars like you did here (e.g. savedIssueType or lastTaskIssue), compared to directly write it like on element:

```
defaultValue={(localStorage.getItem('lastTaskIssue') as ITaskIssue) || null}
```

evereq:
@JoeMakuta honestly, I don't recall where we wanted to use LAST selected task type when we create next task?
We also 100% don't want "Bug" type to be the default btw (you have it in this PR!)...
We talked to load it from the settings, right? I.e. each team may want to define its own default task type! When we seed data in Gauzy, we can set one of the types to be the default type (e.g. say default will be "Story") and that type will be used when new tasks are created, no matter what user selected before.
Also, I recall that @GloireMutaliko21 actually supposed to be working on that... So I don't get what this PR is about and why you worked on it...

JoeMakuta:
> @JoeMakuta honestly, I don't recall where we wanted to use LAST selected task type when we create next task? We also 100% don't want "Bug" type to be the default btw (you have it in this PR!)... We talked to load it from the settings, right? I.e. each team may want to define its own default task type! When we seed data in Gauzy, we can set one of the types to be the default type (e.g. say default will be "Story") and that type will be used when new tasks are created, no matter what user selected before. Also, I recall that @GloireMutaliko21 actually supposed to be working on that... So I don't get what this PR is about and why you worked on it...

GloireMutaliko21:
> @JoeMakuta honestly, I don't recall where we wanted to use LAST selected task type when we create next task? We also 100% don't want "Bug" type to be the default btw (you have it in this PR!)... We talked to load it from the settings, right? I.e. each team may want to define its own default task type! When we seed data in Gauzy, we can set one of the types to be the default type (e.g. say default will be "Story") and that type will be used when new tasks are created, no matter what user selected before. Also, I recall that @GloireMutaliko21 actually supposed to be working on that... So I don't get what this PR is about and why you worked on it...

Sir @evereq I worked on Gauzy API about this task. @JoeMakuta is supposed to work on it for Ever Teams

PR #2545, opened 2024-05-25
Ability to Track time on "Task" page

GloireMutaliko21:
We use usePathname Next.js hook to get the current page. If the current page is the "Task" page, when the user clicks on the start button of the timer, we first set the active task to a displayed task

### DIFF
--- a/apps/web/app/[locale]/task/[id]/page.tsx
+++ b/apps/web/app/[locale]/task/[id]/page.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { useOrganizationTeams, useTeamTasks, useTimer, useUserProfilePage } from '@app/hooks';
+import { useOrganizationTeams, useTeamTasks, useUserProfilePage } from '@app/hooks';
 import { ChildIssueCard } from '@components/pages/task/ChildIssueCard';
 import { RelatedIssueCard } from '@components/pages/task/IssueCard';
 import TaskProperties from '@components/pages/task/TaskProperties';
@@ -25,8 +25,7 @@ const TaskDetails = () => {
 	const router = useRouter();
 	const params = useParams();
 	const { isTrackingEnabled, activeTeam } = useOrganizationTeams();
-	const { getTaskById, setActiveTask, detailedTask: task, getTasksByIdLoading } = useTeamTasks();
-	const { timerStatus } = useTimer();
+	const { getTaskById, detailedTask: task, getTasksByIdLoading } = useTeamTasks();
 	const fullWidth = useRecoilValue(fullWidthState);
 
 	const id = params?.id;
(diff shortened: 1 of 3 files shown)


evereq:
@GloireMutaliko21 I don't get logic you describe...  You talk about individual task page right? I.e. page that display specific task and that we see "timer" component on that page at site header. If timer is counting time for some other task, we should do nothing, unless user stop that timer and start it again! If stop and start again, we need to switch active task for user to be one that s/he currently see on the page! If timer was stopped before and user starting it, then we need to do the same - select active task for the user to be one displayed on this page! Basically any "start" of timer on some task page, should switch user active task to the one that displayed on the page. Is that clear now?

GloireMutaliko21:
> @GloireMutaliko21 I don't get logic you describe... You talk about individual task page right? I.e. page that display specific task and that we see "timer" component on that page at site header. If timer is counting time for some other task, we should do nothing, unless user stop that timer and start it again! If stop and start again, we need to switch active task for user to be one that s/he currently see on the page! If timer was stopped before and user starting it, then we need to do the same - select active task for the user to be one displayed on this page! Basically any "start" of timer on some task page, should switch user active task to the one that displayed on the page. Is that clear now?

Yes sir @evereq It's clear. This is already done in this PR... The only thing I'm not sure should work as such is that currently, when the page for a specific task loads, I check the timer status. If it's already started, nothing happens. If not, I change the user's active task. So what I'd like to know is whether it should work like this (check when the page loads) or whether I should wait for the timer button to be pressed?

evereq:
" If it's already started, nothing happens." - you mean timer continues ticking, but you did not change the active task right?
" If not, I change the user's active task. " - hm, you should ONLY change active task when the user CLICK START on a timer, not when the user just jumps from one task page to another!
How I see it:
- on page load, you only need to make sure the timer "ticking" if it was started before. You don't need to change any active task on the page load event!!
- if the user clicks "Start" button on the timer, you need to change the active task to the one that the user sees on the page
- if the user clicks "Stop" button on the timer, nothing should change in regards to the active task (except I think we have logic to move the task automatically from "In Progress" status to "TODO" (or whatever called the first status of task), but I am not sure if this logic is there already or not, etc (and that logic also affected by setting for team)

Again, honestly, if you think logically about this feature, why we need it!? We want to make sure that user can quickly start tracking time for different tasks right from the tasks pages! I.e. I browsing tasks that exists and see one I like to work on. I click "Start" and timer start for that task and I start working on it. Plus when I moving from one page to another (not only tasks pages, but all app pages), in the header I just want to see my timer status, i.e. if I am running it or not.

GloireMutaliko21:
That's great sir @evereq, the only concern was "when do we change the user's active task". I'll make sure it's when the user starts the timer on that specific task page, not when it loads... Thanks sir

evereq:
@GloireMutaliko21 great, let me know when PR is ready to merge :)

GloireMutaliko21:
Sir @evereq ! The PR is ready to merge :rocket:

PR #2617, opened 2024-06-13
[FIX] Change Validate into Confirm

GedeonTS:


evereq:
@GedeonTS why its not in translations??

GedeonTS:
Thanks for the heads up sir @evereq , here is the translated version:https://github.com/ever-co/ever-teams/pull/2631

Issue #2373, opened 2024-04-07
[Feature]: Task status order ( Kanban + Settings ) 


Anishali2:

rahul-rocket:
@Anishali2 Let's wait to deploy latest changes on API.

https://github.com/ever-co/ever-gauzy/pull/7791

evereq:
Duplicate to https://github.com/ever-co/ever-teams/issues/1374

PR #2651, opened 2024-06-29
[Feature] #2373 Task status order kanban settings

evereq:


evereq:
@Anishali2 can you finish this PR so we can merge it?

PR #2683, opened 2024-07-04
Fix task time tracking issues

GloireMutaliko21:
Fixing update statistics for task time tricking : 
[x] Timer running : When user clicks on Stop timer, the Today/Total worked time are set to new values
[x] Total daily worked hours for user are updated
[x] During running, After each 1 minute, those values are also updated
[x] All tasks that the user worked home appear on Tab "Worked" on the user profile page

:warning: :warning: :warning: Note : WARNING !!! Before merge this PR, please test other parts of code as I commented some critical first loading conditions. Those tests were the main reason of not updating values after pause timer. It took me some time to detect that.

[screencast-localhost_3030-2024_07_04-17_47_41.webm](https://github.com/ever-co/ever-teams/assets/86450367/b5283b0e-e4eb-407e-a1b6-454018ec5992)

### DIFF
--- a/apps/web/app/hooks/features/useTimer.ts
+++ b/apps/web/app/hooks/features/useTimer.ts
@@ -364,8 +364,13 @@ export function useTimer() {
 	}, [timerStatus, setTimerStatus, stopTimerQueryCall, taskId, updateLocalTimerStatus]);
 
 	useEffect(() => {
-		console.log(timerStatus);
-	}, [timerStatus]);
+		if (timerStatus?.running) {
+			const syncTimerInterval = setInterval(() => {
+				syncTimer();
+			}, 60000);
+			return () => clearInterval(syncTimerInterval);
+		}
+	}, [syncTimer, timerStatus]);
 
 	// If active team changes then stop the timer
 	useEffect(() => {
--- a/apps/web/app/hooks/features/useTimer.ts
+++ b/apps/web/app/hooks/features/useTimer.ts
@@ -84,27 +84,27 @@ function useLocalTimeCounter(timerStatus: ITimerStatus | null, activeTeamTask: I
 
 	// Update local time status (storage and store) only when global timerStatus changes
 	useEffect(() => {
-		if (firstLoad) {
-			const localStatus = getLocalCounterStatus();
-			localStatus && setLocalTimerStatus(localStatus);
+		// if (firstLoad) {
+		const localStatus = getLocalCounterStatus();
+		localStatus && setLocalTimerStatus(localStatus);
 
-			const timerStatusDate = timerStatus?.lastLog?.createdAt
-				? moment(timerStatus?.lastLog?.createdAt).unix() * 1000 - timerStatus?.lastLog?.duration
-				: 0;
+		const timerStatusDate = timerStatus?.lastLog?.createdAt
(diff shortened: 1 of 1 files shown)


evereq:
1. "Those tests were the main reason of not updating values after pause timer" - so why we need to remove that!? I.e. if timer is paused what happens now!? Why those conditions were added before and why you think they are needed now? 
2. Also, why refresh every 1 min and not say every 5 seconds? 
3. Is there any risk that this code inside https://github.com/ever-co/ever-teams/pull/2683/files#diff-b60a83ce8f5f5ce9aae8eafd5568f9176a0a5d0c1c556211b467d1708e9aef52R368 runs few times and it schedule FEW times same refresh!? Basically something happens and this code execute again when timer is already running and it schedule SECOND (or more) setTimeout and run again another "refresh" scheduled every 1 min and so on!? We may have that way 100s of refreshes instead of always just one scheduled, no!?

Cedric921 (approved):
Tested on my side, it works well

GloireMutaliko21:
Sir @evereq Those test were to set the local timer on the first loading of the page. These statuses sync the timer duration and allow update tracking statistics for the task.

So we tried first with @Cedric921 by only adding that syncTimer function after pause and still not working as expected unless I comment those If conditions. After commented some of them, testing other parts and seems no code broke. 

"Also, why refresh every 1 min and not say every 5 seconds?" ⇾ It's about making 2–3 requests to API for updating Today work for the task, Total worked and Daily Tracking time total. So I thought it could be better to sync it after 1 min

Cedric921:
> 3\. Is there any risk that this code inside https://github.com/ever-co/ever-teams/pull/2683/files#diff-b60a83ce8f5f5ce9aae8eafd5568f9176a0a5d0c1c556211b467d1708e9aef52R368 runs few times and it schedule FEW times same refresh!? Basically something happens and this code execute again when timer is already running and it schedule SECOND (or more) setTimeout and run again another "refresh" scheduled every 1 min and so on!? We may have that way 100s of refreshes instead of always just one scheduled, no!?

R 2/ Some of these tests were to avoid unexpected behavior during page loading when the timer was running, 
But there is some place where we don't need to check this condition: 
- e.g.: with timerSecondsRef, which updates the progress bar value, we don't need to restrict this function to the first Load only, but we do need to upadte this value every time when var second, as in the useEffect (line 125 to 127).

apps/web/app/hooks/features/useTimer.ts
```diff
@@ -355,11 +355,23 @@ export function useTimer() {
...
+		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, [timerStatus, setTimerStatus, stopTimerQueryCall, taskId, updateLocalTimerStatus]);
 
+	useEffect(() => {
+		if (timerStatus?.running) {
+			const syncTimerInterval = setInterval(() => {
+				syncTimer();
+			}, 60000);
+			return () => clearInterval(syncTimerInterval);
+		}
```
Cedric921, line 372: We need to clear Interval when Component is unmounted

evereq:
@GloireMutaliko21 , ok, I will merge for now, but possible you overlooked some of use cases... we will see during testing..

Issue #2616, opened 2024-06-13
[Feature] 'Daily Plan' | Filter by 'Date range'


KostyaEver:
User should be able to see additional elements for filtering by dates

'Date range' filter should appear only when user sees the next tabs in 'Daily Plans':
- Future
- Past
- All Tasks

<img width="651" alt="Screenshot at Jun 13 11-15-06" src="https://github.com/ever-co/ever-teams/assets/107203576/ddce4b13-b8a0-4f27-a730-f2f23213fbdd">

<img width="666" alt="Screenshot at Jun 13 11-13-45" src="https://github.com/ever-co/ever-teams/assets/107203576/9c1d7796-e875-4421-9c07-0e0bcb0d68a3">

Details:

'Future tasks' should show date ranges  from tomorrow  to the end of the future plan

'Past tasks' should show date ranges from yesterday to the end of the past plan

'All Tasks' should display date ranges all possible periods

PR #2705, opened 2024-07-10
 [Feat] Daily plan date range | Filter 

Innocent-Akim:
Uploading daily_plan_filter.mov…

1.  Daily plan date range | filter  #2616 
Future
Past
All Tasks 

@evereq 
@Cedric921


Cedric921:
https://github.com/user-attachments/assets/2915ed61-c569-4cf2-abbf-5e3af069f708

evereq:
@Innocent-Akim  / @Cedric921 when filter applied, we need some indication about that on the "Filter" button itself (e.g. small number that indicates how many different filters applied right now)

PR #2735, opened 2024-07-15
[Fix] Dailyplan Deleteplan

Innocent-Akim:
Fix: Error on daily plan delete plan

@evereq 
@Cedric921


apps/web/.env copy
```diff
@@ -0,0 +1,112 @@
+RECOIL_DUPLICATE_ATOM_KEY_CHECKING_ENABLED=false
```
evereq, line 1: @Innocent-Akim please remove this copy :)

PR #2737, opened 2024-07-15
[Fix] Daily plan creation popup, filters and sort

GloireMutaliko21:


apps/web/lib/features/daily-plan/create-daily-plan-form-modal.tsx
```diff
@@ -104,14 +104,14 @@ export function CreateDailyPlanFormModal({
 								/>
 							)}
 
-							<InputField
+							{/* <InputField
```
evereq, line 107: @GloireMutaliko21 this supposed to be commented out? Why?

GloireMutaliko21: Sir @evereq I commented this field after long discuss with @KostyaEver. He told me that it not make sense to ask the user how much time he wants to work for future days. I suggested keeping the field but making it not required, but @KostyaEver wanted it to be hidden

evereq: hm, interesting, so that field allowed to enter for any future days how much person think he will work? OR what was idea @GloireMutaliko21

KostyaEver: I confirm that it does not make any sense every time the user adds one task + asks about planned total hours. It's the wrong flow.

Ask about a day planned hours will be in a different way. NOT when a user adds a task to the 'Daily Plan'

@evereq

PR #2747, opened 2024-07-17
fixbug: user should login and create a new team if there is no one

### LINKED ISSUE #2733: [Fix] Login error on Empty Teams for the user
When the user login and does not have any team yet, we should display the home page asking him to create a new team.
The issue is that we get the error described on the below image.

image

CREDO23:
#### Related issues (fixed) ;
- This should also be fixed (when creating a new team)
Screenshot from 2024-07-17 20-49-34

- We should also delete the last selected team from the local storage if the user disposed it


CREDO23:
We should also delete the last selected team from the local storage if the user dispose it

Cedric921 (approved):

PR #2753, opened 2024-07-18
[Bug] Register | Tenant SMTP blocks new users to register (development environment)

GloireMutaliko21:


evereq:
@GloireMutaliko21 I am not sure it's best solution with try and catch, why that method failing for you during development?

GloireMutaliko21:
> @GloireMutaliko21 I am not sure it's best solution with try and catch, why that method failing for you during development?

Sir @evereq The issue was SMTP_USERNAME and SMTP_PASSWORD empty values, now fixed with if condition

PR #2800, opened 2024-07-26
fixbug: blocking any user to start working

### LINKED ISSUE #2757: [Bug] User can NOT start tracking time
The issue is that the user can NOT start tacking Time with the timer.

Screenshot at Jul 18 09-51-00

Screenshot at Jul 18 09-51-32
**On that screen user just simply clicks 'OK' and NOTHING happens, simply can create plan OR NOT**

discussed in this task https://github.com/ever-co/ever-teams/issues/2679
AND here https://github.com/ever-co/ever-teams/issues/2679#issuecomment-2230651487

Expect: if the feature 'Enforse Tracking Time' is OFF - user should be able to see all popups that inform / suggest the user to create a plan / add estimates / provide working hours. But user should be able to ignore / skip it and continue doing what he/she prefers.

CREDO23:
### This PR fixes #2757 

The previous PR that intended to resole the issue had a little problem in the SOFT flow  ⇒  check if a user has already today tasks (which means having the today plan with some tasks added) so that we can show some content in the popup, otherwise show an empty popup (no action to do ⇒ block everything)

#### What I dd
- Remove that check

### DIFF
--- a/apps/web/lib/features/daily-plan/plans-work-time-and-estimate.tsx
+++ b/apps/web/lib/features/daily-plan/plans-work-time-and-estimate.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 import { PiWarningCircleFill } from 'react-icons/pi';
 import { IDailyPlan, ITeamTask, IUser } from '@app/interfaces';
 import { Card, InputField, Modal, Text, VerticalSeparator } from 'lib/components';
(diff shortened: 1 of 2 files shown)


Cedric921 (commented):
After Testing, the Input to "Add planned working hours " is not working
user can't edit the 0 value
Please check and fix @CREDO23

CREDO23:
Okay @Cedric921 ! I just fixws the content in the popup ! 
Let me now, check the input !

Cedric921 (commented):
Check what happens now in the Pop-up:
- Planned working hour now accept negative hours (number): I will create a separated Issue
- Estimated Hours for Task lose focus on  editing mode and make difficult to Add/edit Task Estimated hour

Please see the SS @CREDO23 

https://github.com/user-attachments/assets/274df566-1d4c-41e3-a956-df26a9cc1e5a

CREDO23:
I can see the issue @Cedric921  ! What I have been resolved is the issue that block user to start working (an empty popup). I think you need to create other specific issues instead of fixing unrelated bugs, considering the issue description #2800!

What do you think ? @Cedric921 

However, if you allow, I can solve them in the same PR.

Cedric921:
> I can see the issue @Cedric921 ! What I have been resolving is the issue that block user to start working (an empty popup). I think you need to create other specific issues instead of fixing unrelated bugs in the same PR, considering the issue description #2757
> 
> What do you think ? @Cedric921
> 
> However, if you allow, I can fix them in the same PR.
> 
> PS: The same issues are in production, I did not break an existing feature !

That makes sense @CREDO23 , I've already created a separate Issue.
We can merge this PR @evereq

Cedric921 (approved):

PR #2802, opened 2024-07-26
[Bug] 'Daily Plan' | Action menu displays restricted set of options

Innocent-Akim:
[Fix]: Daily Plan Action Menu Displays

https://github.com/user-attachments/assets/f9e36017-2bd3-4902-bbfe-096c0fedd561


Cedric921 (approved):
On This Tab, we will need a small Toast to notify whether or not the task has been added to the plan (Today/Tomorrow/ ...).

PR #2806, opened 2024-07-29
[Feat] Enforced plan should let track time only on planned task

### LINKED ISSUE #2785: [Feature] 'Daily Plan' | Enforced Plan - should let track time ONLY on planned task
If in Settings set 'Required Plan'

Screenshot at Jul 23 08-54-15

The system should control on which tasks the user is tracking time.

The system should Enforce to track time ONLY on tasks that planned to that Date

In case the user selects another task and clicks 'Track' there should be a popup that suggests adding the task to the Plan.

text: 'This task was not planned for Today's plan. Would you like to add to the plan?' Yes/No

in case 
'Yes' - simply add the task to the plan
'No' - remain the same case (can track time only for planned tasks)

**IMPORTANT**: it should work ONLY when the Plan is Enfoced for the user, if it is SOFT mode (when Enforced is OFF) User should be able to track time for ANY task

CREDO23:
### This PR fixes #2785 

Here is a short video [Loom](https://www.loom.com/share/21755691bef44b7992acb4680c5f876f?sid=2c2a69b6-fef2-424a-acea-a1b35f41b7f9)


Cedric921 (changes requested):
This feature should be available wherever there's a Timer Start/Stop button, whether the Team is using Enforced Plan mode.

Now :
- With The Timer Button on the home page works fine (ActiveTask).
- The issue is with the Profile page, Assigned Tasks, Start Timer for no planned Tasks

https://github.com/user-attachments/assets/b2038923-5aca-4529-954b-8fd797500bfe

evereq:
@CREDO23 / @Cedric921 please fix merge conflicts carefully and work toward making this PR mergable.

CREDO23:
> This feature should be available wherever there's a Timer Start/Stop button, whether the Team is using Enforced Plan mode.
> 
> Now :
> 
> * With The Timer Button on the home page works fine (ActiveTask).
> * The issue is with the Profile page, Assigned Tasks, Start Timer for no planned Tasks
> 
>  Screen.Recording.2024-07-29.at.23.41.53.mov

@Cedric921, I think it's better to merge these changes first !

The problem is that there is a code duplication for each start/stop timer button in the app, I can create an issue and open another PR to fix it !

What do you think ? @Cedric921 @evereq

evereq:
"The problem is that there is a code duplication for each start/stop timer button in the app, I can create an issue and open another PR to fix it !" - yes please @CREDO23

PR #2858, opened 2024-08-07
fix: popup should not appear when the user clicks 'Back' (or the profile page) button

### LINKED ISSUE #2614: [Feature] 'Daily Plan' compare Estimated tasks time and Planned work hours
WORK ON THE TASK ONLY AFTER CONFIRMATION

Right after SOFT Plan FLOW was implemented (https://github.com/ever-co/ever-teams/issues/2680) we need to improve SMART approach to the popup that decides if the provided 'Planned time' & 'Planned task estimations' are balanced well.

The main trigger if that moment is when the system asks to provide 'Planned time' for Today

The system should check 

1. if user has 'Today' plan or not (If has) (implemented)
2. it checks if all tasks have Estimates (if no, ask to add in a popup)  (implemented)
3. it checks if 'Planned work time' was provided (if no, ask to add in a popup)  (implemented)
**4. The system should check if the Total Estimates of all planned tasks are approx the same as 'Planned work hours''**

It should display a NEW popup in case
1. User added too many tasks with total estimates that are more than 'Planned work hours' (afford +1h to planned work time)
2. User added a few tasks with total estimates that are much less than 'Planned work hours' (afford -1h to planned work time)

So the range can vary from -1 to +1 hour from the defined value 

The popup should be similar to described here https://github.com/ever-co/ever-teams/issues/2603

just add one more line under the task list of the popup in RED (as a Notification)

'Please, correct one of the more appropriate values
- 'Planned work hours'
- 'Reestimate task(s)'
- 'Replan task' OR 'Plan task' (depends on whether we need to remove or add more tasks to fulfill the work day)

**Note**: so the user can see https://github.com/ever-co/ever-teams/issues/2603 with additional information so that we don't need to design more popups and we should not force a user to make all of these corrections. Just to inform the user about that difference for V.1

Screenshot at Jun 13 17-33-33

### LINKED ISSUE #2849: [Bug] 'Daily Plan' | unexpected Popup is popping up
The bug randomly happens when the user is trying to go back to the main page from the profile page

Case how it might be reproduced:
- Add task to 'Today plan'
- Click 'Back'
- the popup should appear (that should NOT happen in such case)

**Expect**: popup should not appear when the user clicks 'Back' (or the profile page) button

Screenshot at Aug 07 09-38-58

What appears as the result
Screenshot at Aug 07 09-38-27

### LINKED ISSUE #2913: [Bug] 'Planned' | Popup 'Planned' triggered by wrong event
The issue is that the 'Planned' popup appears right away user logs in
It happens because 'Today' user has Planned tasks.

**Expect**: When 'SOFT' flow is ON, and the user has planned tasks for Today, it should NOT call the popup AFTER the user logged in.

The popup should be called ONLY if the user clicks the button 'Start tracking' time the user has selected the planned task (on the Dashboard page) 

- Planned Time was not provided
- One or few tasks are not estimated
- Total Estimated Time & Planned work time have difference more that -/+ 1 hour

**Note**: if the selected task was not planned Today OR No plan at all, the system calls other popups.

Example: Now after login when there is a Today's Plan 
Screenshot at Aug 16 09-20-50

Screenshot at Aug 16 09-21-33

Screenshot at Aug 16 09-22-04

CREDO23:


evereq:
@CREDO23 be aware of merge conflict

PR #4041, opened 2025-10-15
[FixBug] Profile Page | Plans | “Planned Time” input hides value in edit mode

CREDO23:
## Description

- Fix plan header | planned time input in the profile page
- [JIRA Ticket](https://evertech.atlassian.net/browse/ETP-130?atlOrigin=eyJpIjoiM2M4ZDcwZGVlYWQyNGVjM2IyZmEzN2U3MzBkOWJkYjEiLCJwIjoiaiJ9)

## Previous behavior

https://github.com/user-attachments/assets/245174c2-9153-4e40-84a9-0a5632fdd0d8

## Current behavior

https://github.com/user-attachments/assets/69e1d4a1-8929-4681-bc74-da6ea5ce5fd3

## Type of Change

- [x] Bug fix (fixes a problem)
- [ ] Documentation update

## ✅ Checklist

- [ ] I updated or created related documentation if needed

## Summary by CodeRabbit

- New Features
  - Editable planned-time input with H:MM and decimal parsing, save flow, and success/error toasts.
  - Period selector added to Auto-Close Issue and Auto-Archive Issue settings.

- Bug Fixes
  - Time field pre-fills from plan, validates input, avoids no-op updates, and restores state on failure.

- Style
  - Cleaned up layout and spacing in Team Issue Settings to better accommodate new controls.

### DIFF
--- a/apps/web/core/components/daily-plan/plan-header.tsx
+++ b/apps/web/core/components/daily-plan/plan-header.tsx
@@ -126,7 +126,7 @@ export function PlanHeader({ plan, planMode }: { plan: TDailyPlan; planMode: Fil
 					<div className="flex">
 						<input
 							value={inputValue}
-							type="string"
+							type="text"
 							className={clsxm(
 								'p-0 text-xs font-medium text-center bg-transparent border-b outline-none max-w-[54px]'
 							)}
--- a/apps/web/core/lib/helpers/date-and-time.ts
+++ b/apps/web/core/lib/helpers/date-and-time.ts
@@ -326,7 +326,7 @@ export function parseStringInputToHours(input: string): { hours?: number; error?
 
 	// Support decimal format (4.5, 3.25, etc.)
 	const decimal = Number(raw.replace(',', '.'));
-	if (!isNaN(decimal)) {
+	if (!isNaN(decimal) && decimal >= 0) {
 		return { hours: decimal };
 	}
 


apps/web/core/components/daily-plan/plan-header.tsx
```diff
@@ -82,11 +82,12 @@ export function PlanHeader({ plan, planMode }: { plan: TDailyPlan; planMode: Fil
 					<div className="flex">
 						<input
 							min={0}
+							value={time}
 							type="number"
 							className={clsxm(
 								'p-0 text-xs font-medium text-center bg-transparent border-b outline-none max-w-[54px]'
 							)}
-							onChange={(e) => setTime(parseFloat(e.target.value))}
+							onChange={(e) => setTime(Number(e.target.value))}
```
evereq: @CREDO23 any reason you use `Number` here, even according to AI suggestion: "The fix attempts to solve the visibility issue but introduces a critical bug where parseFloat(formatIntegerToHour()) loses decimal precision. For example, 5.5 hours becomes 5 hours. This will cause incorrect time values to be displayed and saved.". Are you reviewing suggestions by AI (not just comments, but BODY of message from it)?

CREDO23: Yes @evereq, here we need to keep the decimal precision of `workedTime`. But previously we had `parseFloat` which was problematic . Now i changed it to be `Number` so that we can keep all decimal precision .

This is the main reason I changed it to be `Number` instead of `parseFloat`

evereq: yessss @CREDO23 , but AI suggested that it will anyway lost values like "5:30" !! So it's like "5.5" will work as before, but 5:30 will not work... We need to make sure user can enter both 5.5 and 5:30 and we correctly convert etc. Try to run GPT, you will see it suggest a bit more sofisticated code like
```
function parseHourString(str: string): number {
  if (!str.includes(":")) return Number(str);
  const [hours, minutes] = str.split(":").map(Number);
  return hours + minutes / 60;
}

// usage:
onChange={(e) => setTime(parseHourString(e.target.value)));
```
So if you do want to change that line, maybe let's change it to make sure it fully works?

CREDO23: I see @evereq, this is something new ! And yes, we can make it work like that ...

Actually, previously, user could only enter hours ..... e.g : 2, 3, 4, 5.2, 5.3 and directly send to the backend

Now, if we want, we can also support input like 5:30 (hours : minutes) ! But this wasn't supported !

CREDO23: Let me know and I will implement it @evereq

CREDO23: Hi @evereq, I have updated it … Now the user can type in `2:30` format or `4.5` format


https://github.com/user-attachments/assets/ccb3134b-7da0-436b-8fb1-f79bde26dee3

PR #4097, opened 2025-10-27
[Bug] Missing save/confirm action for Team Color & Team Emoji inputs

CREDO23:
## Description

- Fix team settings color and emoji inputs
- [JIRA Ticket](https://evertech.atlassian.net/browse/ETP-143)

## Current behavior

https://github.com/user-attachments/assets/ccb27ab6-e7d2-4714-8164-45c9a2cbd8eb

## Type of Change

- [x] Bug fix (fixes a problem)
- [ ] Documentation update

## ✅ Checklist

- [ ] I updated or created related documentation if needed

## Summary by CodeRabbit

* **New Features**
  * Color and emoji pickers now include an explicit Save action with loading and saved-state indicators; edits are persisted only when saved.
  * Pickers disable editing while a save is in progress and show clear visual feedback.

* **Style**
  * Minor spacing adjustment in the team management popover for a tighter layout.

### DIFF
--- a/apps/web/core/components/common/color-picker.tsx
+++ b/apps/web/core/components/common/color-picker.tsx
@@ -30,7 +30,7 @@ export const ColorPicker = ({
 	const buttonRef = useRef<HTMLButtonElement>(null);
 	const panelRef = useRef<HTMLDivElement>(null);
 	const [disabled, setDisabled] = useState<boolean>(true);
-	const [isInternalUpdate, setIsInternalUpdate] = useState(false);
+	const [,setIsInternalUpdate] = useState(false);
 	const [loading, setLoading] = useState(false);
 
 	const toggleDisabled = useCallback(() => {
@@ -140,7 +140,7 @@ export const ColorPicker = ({
 							{isTeamManager && (
 								<div className="flex mr-[0.5rem] gap-3">
 									{open ? (
-										<button onClick={handleSave}>
+										<button type="button" onClick={handleSave}>
 											{loading ? (
 												<LoaderCircle
 													className="w-[18px] h-[18px] animate-spin"
@@ -163,7 +163,7 @@ export const ColorPicker = ({
 											) : (
 												<>
 													<button
-														disabled={!isTeamManager || loading}
+														disabled={loading}
 														className={`z-50 outline-none`}
 														onClick={() => {
 															setDisabled(!disabled);
--- a/apps/web/core/components/common/emoji-picker.tsx
+++ b/apps/web/core/components/common/emoji-picker.tsx
@@ -126,7 +126,7 @@ export const EmojiPicker = ({
 											) : (
 												<>
 													<button
-														disabled={!isTeamManager || loading}
+														disabled={loading}
 														className={`z-50 outline-none`}
 														onClick={() => {
 															setDisabled(!disabled);


CREDO23:
> @CREDO23 this PR removes some logic, e.g. isTeamManager and some other things and I can't easy trace WHY it does that if title of PR is "missing save/confirm action"... I think @NdekoCode have to review PR more deeply, but I wonder if you can add COMMENTS in your PR (in GitHub) and explain WHY you removing some things / why they are not needed etc. Thanks

- `isManager` was handled inside the components (color picker / emoji picker) to disable the component for non MANAGER role, that should not live inside the components because not everywhere we will need that logic : that logic needs to live in the parent component !, i.e. it should be a reusable component, so I put it out of that component and use `disable` props to manager when that component should be disabled
- I also migrate to `headless/ui` popover for simplicity
- Remove unnecessary side effects (manually adding / removing outside clicks) because headless/ui popover manage them out of the box

here are some reasons about what was changed @evereq / @NdekoCode

apps/web/core/components/common/color-picker.tsx
NdekoCode, line 1: We lost this ui @CREDO23 

Image

NdekoCode, line 1: Now, `onChange` is still present, but the component also call `editOrganizationTeam` internally.
That means both can trigger updates, leading to double updates or inconsistent state.

apps/web/core/components/common/emoji-picker.tsx
NdekoCode, line 1: `EmojiPicker` calls both:
```tsx
onChange(emoji.native)
await editOrganizationTeam({ emoji: value.native })

```
This can result in local and remote states diverging if the parent updates asynchronously.

apps/web/core/components/common/emoji-picker.tsx
```diff
@@ -2,139 +2,151 @@ import data from '@emoji-mart/data';
...
+														e.stopPropagation();
+														await removeTeamEmoji();
+													}}
+												/>
+											)
+										) : null}
+									</div>
+								)}
+							</PopoverButton>
+							<PopoverPanel anchor="bottom" className="shadow-lgcard mt-6">
```
NdekoCode: this must be `shadow-lg card` instead of `shadow-lgcard`

apps/web/core/components/common/color-picker.tsx
NdekoCode, line 1: What do you think about of instead of removing that `isTeamManager` we could do something like ?:
```tsx
<ColorPicker
  defaultColor={...}
  onChange={...}
  disabled={!isTeamManager}
  requireManagerRole // <— optional prop for clarity
/>
```
This would maintain reusability while making the intention explicit.

NdekoCode (commented):
@CREDO23 Moving the `isTeamManager` logic outside the components is a good step toward making them reusable and context-independent.

But there is an inconsistency, previously, the parent (`TeamSettingForm`) handled data synchronization via onChange and submitted all updates together.

Now, `ColorPicker` and `EmojiPicker` call `editOrganizationTeam` directly inside themselves.

This introduces side effects within the UI components, which breaks the separation between presentation and data management.

I think we can keep API calls in the parent (`TeamSettingForm`) and let the components emit only onChange events.
If inline updates are desired, use an explicit prop such as:
```tsx

<ColorPicker autoSave onChange={...} />

```

Also the previous version had clear control over disabled and editing state (for example toggled with `EditPenUnderlineIcon`).
Now, in `ColorPicker` and `EmojiPicker`, editing is implied by the `Popover` state only.

we have to conside re-adding a small “edit mode” toggle for clarity, or at least ensure that the popover visually indicates active edit mode.

I believe we can standardize the UX behavior across all editable fields because `TeamSize` still uses the “save/cancel” modal pattern, while `ColorPicker` and `EmojiPicker` now self-update immediately.

apps/web/core/components/common/emoji-picker.tsx
CREDO23: I used shadow-lgcard to add shadow, it is not a typo, it is a valid tailwindcss class

apps/web/core/components/common/color-picker.tsx
CREDO23: It depends...., we can add it, but for now I think `disable` is enough

apps/web/core/components/common/emoji-picker.tsx
CREDO23: I do not get this please

apps/web/core/components/common/color-picker.tsx
NdekoCode: Did you checked this @CREDO23  ?

apps/web/core/components/common/emoji-picker.tsx
NdekoCode: `shadow-lgcard` is not a valid tailwindcss class @CREDO23 https://github.com/ever-co/ever-teams/pull/4097#discussion_r2472726984

CREDO23: it is configured in our tailwind config @NdekoCode , please check

PR #4109, opened 2025-11-04
fix: server web api url setting not applying on runtime process

syns2191:
# 🚀 Pull Request Title

_A short and clear title that describes what this PR does._

Example:
> Add Retry, Cache, and Logging Features to API Service

## Description

Please describe **what you did**, and **why**.

- What problem or feature does this PR address?
- What changes were made?
- Why are these changes useful?

Example:
> This PR adds advanced features to the API request system:
>
> - Retry failed requests with exponential delay
> - Cache GET responses with TTL
> - Log all requests and responses
> - Allow cancelling requests
>
> These features improve reliability, performance, and debugging.

### Major Changes

Example:
> Here are the major changes in that his PR adds
>
> - New `CookieManagement` class for cookie handling
> - Cancel requests using `AbortController`
> - Automatically inject tenant and organization headers

### Minor Changes

Example:
> Here are the minor changes in that his PR adds
>
> - Update `tast-status`  component style
> - Work on the theme toggler
> - Remove unused imports

## How to Test This PR

Please explain clearly how to test the changes locally:

Example:
>
> 1. Run the app with `yarn web:dev`
> 2. Open the browser at `http://localhost:3030`
> 3. Try navigating to a page that uses API calls (e.g., `/tasks`)
> 4. Check:
   >
   > - Logs in the console
   > - Retry works on failed requests
   > - Cache works on repeated GET requests
   > - Errors are handled properly
   > - Cancelling navigation cancels requests
   > - Remove unused imports

> If this PR affects UI:
>
> - Include before and after screenshots
> - Explain any design or UX changes

## Screenshots (if needed)

| Before           | After            |
| ---------------- | ---------------- |
| (Add screenshot) | (Add screenshot) |

> You can also add videos or logs here.

### Previous screenshots

Please add here videos or images of the previous status

### Current screenshots

Please add here videos or images of the current (new) status

## Related Issues

Please list related issues, tasks or discussions:

Example:
>
> - Closes #12343434344
> - Related to #567834232

## Type of Change

- [ ] Bug fix (fixes a problem)
- [ ] Documentation update

## ✅ Checklist

Please confirm you did the following before asking for review:

- [x] I updated or created related documentation if needed

## Notes for the Reviewer (Optional)

_Add here any context, help, or known issues for the person reviewing:_

- Example: “The retry logic uses `setTimeout` for now – may need refinement.”
- Example: “PostHog is disabled in dev, enable it in `.env` to test logs.”

## ⚠️⚠️⚠️ Reviewers Suggested

- `@evereq` for architecture validation
- `@ndekocode` for integration review
- `@Innocent-Akim` for auth and cookie handling and assistance
- `@AnicetFantomas` and `@Sergemuhundu` for mobile app and some web issues
- `@Cedric921` and `@GloireMutaliko21` for complex issues

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved error content while improving error throw behavior to avoid altering control flow.

* **Improvements**
  * Simplified desktop API URL resolution and reduced desktop-specific override logic.
  * Deferred desktop environment loading to an async path when running as a desktop app to better surface runtime server URLs.
  * Build script updated to mark desktop builds so runtime desktop behavior is enabled.

### DIFF
--- a/apps/web/core/services/client/axios.ts
+++ b/apps/web/core/services/client/axios.ts
@@ -1,8 +1,7 @@
 import {
 	APPLICATION_LANGUAGES_CODE,
 	DEFAULT_APP_PATH,
-	GAUZY_API_BASE_SERVER_URL,
-	IS_DESKTOP_APP
+	GAUZY_API_BASE_SERVER_URL
 } from '@/core/constants/config/constants';
 import { getAccessTokenCookie, getOrganizationIdCookie, getTenantIdCookie } from '@/core/lib/helpers/cookies';
 import { AxiosRequestConfig, AxiosResponse } from 'axios';


apps/web/core/services/client/axios.ts
```diff
@@ -86,17 +86,17 @@ export const getAPIDirect = async (): Promise<APIService> => {
...
-	if (typeof window !== 'undefined') {
-		try {
-			const serverConfig = await api.get<{ NEXT_PUBLIC_GAUZY_API_SERVER_URL: string }>('/desktop-server');
-
-			return serverConfig?.data?.NEXT_PUBLIC_GAUZY_API_SERVER_URL;
-		} catch (error) {
-			return GAUZY_API_BASE_SERVER_URL;
+export async function desktopServerOverride(): Promise<string> {
+	try {
+		const resp = await fetch('/api/desktop-server', { cache: 'no-cache' });
```
evereq: @syns2191 I don't get, why we need to do an HTTP call here? You do realize that making such a call will be slow even if success and it will take much longer if it fail? So the caller to this function desktopServerOverride() will have to WAIT in any case! You use this function only in Desktop app or Web app also make a call to it!? If Web app also using it, now we can see a reason of delays in calls to APIs from web etc... Can you explain more about what is going on here?

CC: @NdekoCode can you review this code please? I am VERY worry about having such a function (before and after changes in this PR) that making a calls to some API endpoint and in `catch` statement we return some URL value etc. Calls to such function will slow down anything!!!!

syns2191: @evereq, this case is only when in desktop/electron mode. i think we need this to applying the setting that allowing change the GAUZY_API_SERVER_URL

evereq: @syns2191 how frequently it get's called? Is it like one time when app start or it happens many time when user doing some actions etc? I do NOT like this code at all, especially in the apps/web/core!!!

syns2191: It should be every time  when web try to access the API. Yes, it is bad, I think. It's okay if still use this way, but just do it one time when the web is reloading, or do you want me to completely change the way?
@evereq

.scripts/configure.electron.ts
```diff
@@ -34,7 +34,7 @@ function updateWebConstant(setDesktopApp: boolean) {
 		const [from, to] = setDesktopApp ? [envCheck, hardcoded] : [hardcoded, envCheck];
 
 		if (!fileContent.includes(from)) {
-			throw new Error(`Expected content not found in ${filePath}`);
+			console.error(`Expected content not found in ${filePath}`);
```
NdekoCode: @syns2191 maybe we can keep that `throw new Error(`Expected content not found in ${filePath}`);` here

PR #4423, opened 2026-08-15
ci: add external uptime monitoring (GitHub-hosted vantage point outside the homelab)

evereq:
# 🚀 Add external uptime monitoring (GitHub-hosted vantage point)

## Description

**The problem this solves.** Every probe we currently have for Ever Teams — Kubernetes
liveness/readiness probes, Uptime Kuma, the Prometheus blackbox exporter — runs *inside* the
homelab cluster that also owns our egress uplink. They all share a single network fate, so they
cannot answer the one question that matters during an incident: **"is the platform actually down,
or can *we* just not reach it?"**

On **2026-08-15** a routing fault at the ISP (Orange, Spain) made Cloudflare's `188.114.96.0/22`
prefix unreachable **from the homelab only**. `ever.team`, `app.ever.team` and `api.ever.team` all
resolve into that exact prefix:

```
ever.team       A  188.114.96.5, 188.114.97.5
app.ever.team   A  188.114.96.5, 188.114.97.5
api.ever.team   A  188.114.97.5, 188.114.96.5
```

Every internal probe went red at once and it looked precisely like a production outage. It was
not — real users on the public internet were served normally the entire time. Hours were burned
on a false alarm.

This PR adds the missing **independent, outside-the-homelab vantage point**.

## What Was Changed

Exactly **one new file**: `.github/workflows/uptime-external.yml`. Nothing else is touched — no
existing workflow, no release/desktop/mobile-publish pipeline is modified or triggered.

### Major Changes

- **Runs on `ubuntu-latest` (GitHub-hosted) by design.** This is a deliberate, documented
  exception to this repo's standing "self-hosted runners only" CI policy, and the file carries a
  large banner comment explaining why. Our self-hosted ARC runners live in the same homelab,
  behind the same uplink, in the same failure domain as the services being probed — putting this
  workflow on them would silently destroy its only value and would fail us in exactly the incident
  it exists to catch.
- **Triggers:** `schedule` (`*/15 * * * *`) plus `workflow_dispatch` for on-demand checks.
- **Asserts on body content, not just HTTP 200.**
- **Records the full curl timing breakdown** — `time_namelookup`, `time_connect`,
  `time_appconnect`, `time_starttransfer`, `http_code`, plus the peer IP actually reached. This is
  the signal that resolved the 2026-08-15 incident: a *successful* DNS lookup with `time_connect`
  pinned at `0.000000` proves a TCP-level routing blackhole, whereas a merely broken application
  still completes the handshake and answers with a 5xx.
- **3 consecutive failures, 20s apart, before anything alerts** — a single transient blip cannot
  raise an issue.
- **Known-good control endpoint checked first.** If our endpoints look down *and* the control is
  also unreachable, the fault is the GitHub runner's own network, so the run refuses to file an
  alert and says so loudly. A check with no control is how you manufacture a confident wrong
  answer.
- **No issue spam.** One long-lived issue per incident, labelled `uptime-alert`: the first failure
  opens it, subsequent failing runs *comment* on the same issue, and recovery comments and closes
  it. Issue spam is the main reason monitoring like this gets muted and then deleted.

### Minor Changes

- Least-privilege `permissions:` — `issues: write`, `contents: read`. No new secrets: the
  automatic `GITHUB_TOKEN` is the only credential used.
- No `actions/checkout` — the job needs nothing from the tree, which keeps a healthy run to a few
  seconds.
- `concurrency` group so two runs can never race over the same alert issue.
- Job summary carries the same report as the issue body, including a short "how to read this"
  note on the timing columns.

## Endpoints monitored (and how they were verified)

Each was verified against the live endpoint **before** being hard-coded — nothing here was
guessed. Because the homelab uplink is currently blackholing that Cloudflare prefix, verification
was done through an external relay so the responses came from outside this network.

| Endpoint | Assertion | Verified response |
| --- | --- | --- |
| `https://app.ever.team/api/health` | body contains `Ever Teams Next.js API` | `{"data":{"status":200,"message":"Ever Teams Next.js API"},"response":{"url":"https://app.ever.team/api/"}}` |
| `https://api.ever.team/api/health` | body contains `"status":"ok"` | `{"status":"ok","info":{"database":{"status":"up"},"storage":{"status":"up"},"cache":{"status":"up","cacheType":"redis"},"redis":{"status":"up"}},"error":{},...}` |
| `https://ever.team/` | HTTP 200 + body ≥ 1024 B | server-rendered marketing site, no health route, so a non-trivial body is the assertion |

Notes:

- `app.ever.team/api/health` is the Next.js route at `apps/web/app/api/health/route.ts` — the same
  path the Kubernetes readiness/liveness probes use in `.deploy/ever-k8s/ever-teams-web.yaml`.
- `api.ever.team` is the Gauzy headless API this app runs on
  (`GAUZY_API_SERVER_URL` in `apps/web/.env.sample`). Its NestJS Terminus health route reports
  database/storage/cache/redis, so asserting `"status":"ok"` also catches a booted-but-degraded
  API whose database or Redis dependency has failed.
- **Deliberately not monitored:** `stage.ever.team` and `demo.ever.team`. Both are healthy and
  return the same marker, but they are WIP environments where short outages are expected and
  tolerated; alerting on them would train everyone to ignore the label. They are one line each to
  add if that changes.

## Cost

**$0.** `ever-co/ever-teams` is a **public** repository, and GitHub-hosted runner minutes are free
and unlimited on public repos — the 2,000 min/month quota only applies to private ones. That is
why the 15-minute cadence is affordable here.

For reference if the repo ever goes private: a healthy run is a handful of curls (~15 s, billed as
1 min), so `*/15` would be roughly **2,880 min/month** — over the free quota. The file carries a
comment saying to drop to `0 * * * *` (~720 min/month) in that case.

## How to Test This PR

1. **Manual run** (after merge, since `workflow_dispatch` only becomes available once the workflow
   is on the default branch):
   `gh workflow run uptime-external.yml --repo ever-co/ever-teams`
2. Check the run's **job summary** for the timing table.
3. Confirm `runs-on` resolved to a GitHub-hosted runner and the reported public egress IP is a
   GitHub IP, not ours.

### Verification already performed

- **YAML parses** and registers one job with five steps; `runs-on` is the literal `ubuntu-latest`;
  triggers, `permissions` and `concurrency` all resolve as intended.
- **No bare `${{ }}` expression anywhere in the file** (including in comments) — that silently
  invalidates a whole workflow, registering zero jobs. Explicitly grepped for; the only
  expressions present are `${{ secrets.GITHUB_TOKEN }}`, `${{ github.repository }}` and
  `${{ steps.probe.outputs.egress_ip }}`.
- **Every `run:` block passes `bash -n`.**
- **The probe script was actually executed**, not just eyeballed:
  - *Failure path* — run from the affected network, it correctly declared all three endpoints down
    after 3 spaced attempts and produced exactly the intended diagnostic signal
    (`dns=0.008118s`, `connect=0.000000s`, `http=000`), while the control endpoint passed in the
    same run (`connect=0.055973s`) — i.e. it correctly identified "the targets are unreachable
    from here, but this host's network is otherwise fine". Total runtime 3 m 39 s, inside the
    10-minute job timeout.
  - *Success path* — with the same assertions routed through a reachable relay, all three passed
    on the first attempt and the run reported `failed=false` in ~15 s.
  - *Negative control* — with a deliberately bogus body marker and an impossible size threshold,
    the affected endpoints failed with the correct reasons while the untouched endpoint still
    passed, proving the assertions genuinely discriminate rather than always passing.
- **`gh issue list --label uptime-alert`** returns exit 0 and empty output while the label does not
  yet exist, so the recovery step cannot break on a fresh repo.
- **Repo/org default workflow permissions are `write`**, so the least-privilege `issues: write`
  request is grantable; Issues are enabled on the repo.
- **Spellcheck**: the file passes this repo's strict `cspell` gate (`Issues found: 0`), via inline
  `cspell:ignore` directives so `.cspell.json` did not need modifying.
- **LF line endings**, no CRLF.

## Alerting behaviour

| Situation | Behaviour |
| --- | --- |
| All endpoints healthy | Run is green. If an `uptime-alert` issue is open, it gets a recovery comment and is closed. Otherwise nothing happens. |
| 1–2 transient failures | **No alert.** The endpoint must fail all 3 spaced attempts. |
| 3 consecutive failures, control OK | Opens an `uptime-alert` issue (or comments on the existing open one) with the failing URL, HTTP code, full timing breakdown and the runner's public egress IP. Run goes red, which also emails the repo owner. |
| 3 consecutive failures, control also down | **No issue filed.** Logs an explicit error that the runner's own network is suspect and these results cannot distinguish an outage from a GitHub network fault. |
| Recovery | Comments the passing report on the open issue and closes it as completed. A new issue is opened if it recurs. |

## Related Issues

Follow-up from the 2026-08-15 false-alarm incident. Sibling PRs adding the identical workflow are
being opened against `ever-co/ever-gauzy` and `ever-works/ever-works`.

## Type of Change

- [x] New feature (adds functionality) — CI/observability only, no application code
- [ ] Bug fix (fixes a problem)
- [ ] Documentation update

## ✅ Checklist

- [x] I tested my changes locally (see Verification above)
- [x] I updated or created related documentation if needed (the rationale lives in the workflow
      file itself, where a future maintainer will actually see it)

## Notes for the Reviewer

- **The single most important review point:** please do not "fix" `runs-on: ubuntu-latest` to a
  self-hosted/ARC runner. That would put the monitor back inside the failure domain it exists to
  see past, and it would go blind in precisely the incident it was written for. The file has a
  banner comment saying so, but it is worth a human sign-off.
- This changes nothing about the application, the build, or any deploy path. It adds no
  dependency, no secret, and touches no existing workflow. `web.before-merge.yml` is path-filtered
  and will not trigger on this file.
- `workflow_dispatch` will not appear in the Actions UI until this is merged to `develop` (the
  default branch) — GitHub only exposes manual and scheduled triggers from the default branch.
  Same for the schedule: it starts running once merged.
- Cadence and endpoint list are the two knobs most likely to need tuning; both are single-line
  edits at the top of the file and in the `ENDPOINTS` block.

## ⚠️ Reviewers Suggested

- `@evereq` for the runner-policy exception and cadence sign-off

---
## Summary by cubic
Adds `.github/workflows/uptime-external.yml` to monitor `ever.team`, `app.ever.team`, and `api.ever.team` from GitHub-hosted runners so homelab/ISP faults don’t read as platform outages. Previously all probes ran inside the homelab; now an external vantage point verifies availability and files actionable alerts.

- Triggers every 15 minutes and via `workflow_dispatch`; runs on `ubuntu-latest` only by design; requires 3 consecutive failures 20s apart; job timeout 15 minutes.
- Asserts on body markers and minimum size; `https://ever.team/` now checks a semantic marker `name="author" content="Ever Teams"` (not a title string) to avoid false UP/DOWN on interstitials or rebrands.
- Control is `https://checkip.amazonaws.com`; if control fails, the run is INCONCLUSIVE and no issue is filed; control status is included in the timing table.
- Records curl timings and exit codes; captures and validates the runner’s public egress IP.
- Alerting reuses a single `uptime-alert` issue (exact-title match; searches up to 100 open issues to avoid truncation) and closes it on recovery; emits an error only when endpoints failed and control is down, and a warning when only control flaps on an otherwise healthy run.
- Permissions: `issues: write`, `contents: read`; uses `GITHUB_TOKEN`; no secrets; no `actions/checkout`; no other files or pipelines change; $0 on this public repo.

- Required for reviewers: do not change `runs-on: ubuntu-latest` to self-hosted/ARC — that would remove the independent vantage point.

<sup>Written for commit 665e6469689b2c5dad33ccecded1c84b7a16ca82. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ever-co/ever-teams/pull/4423?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added automated monitoring of public service endpoints every 15 minutes.
  * Checks cover availability, connectivity, security, response timing, and transfer health.
  * Confirmed outages generate tracked alerts, with recovery updates and automatic closure.
  * Diagnostic summaries and supporting data are retained for service investigations.

* **Reliability**
  * Monitoring failures are verified against an independent control service to reduce false alerts.
  * Inconclusive and control-service-only failures are reported separately from confirmed outages.

### DIFF
--- a/.github/workflows/uptime-external.yml
+++ b/.github/workflows/uptime-external.yml
@@ -231,8 +231,12 @@ jobs:
           #                               "status":"ok" also catches a booted-but-degraded API
           #                               whose database or Redis dependency has failed.
           #   ever.team/               -> the marketing site, server-rendered HTML. No health
-          #                               route exists, so we assert 200 + a body marker taken
-          #                               from the <title>. A size floor alone was NOT enough:
+          #                               route exists, so we assert 200 + a SEMANTIC marker: the
+          #                               `<meta name="author" content="Ever Teams">` tag. NOT the
+          #                               <title> tagline -- pinning health to marketing copy means
+          #                               a rebrand takes the site "DOWN" on a perfectly healthy
+          #                               deploy, which is the false-alarm class this workflow
+          #                               exists to end. A size floor alone was NOT enough:
           #                               the real page is ~59,796 B against a 1024 B floor, so a
           #                               Cloudflare interstitial, an "Always Online" cached
           #                               shell, a Next.js error page or a wrong-environment
@@ -261,7 +265,7 @@ jobs:
           done <<'ENDPOINTS'
           https://app.ever.team/api/health|Ever Teams Next.js API|10
           https://api.ever.team/api/health|"status":"ok"|10
-          https://ever.team/|Time Tracking Platform|1024
+          https://ever.team/|name="author" content="Ever Teams"|1024
           ENDPOINTS
 
           # -------------------------------------------------------------------------------
@@ -348,8 +352,16 @@ jobs:
           # Match the exact title this workflow writes, not the label alone. The recovery step
           # does not merely comment -- it CLOSES. A human who hand-labels a related incident
           # issue must never have it silently auto-closed by the monitor.
-          EXISTING="$(gh issue list --label "$LABEL" --state open --limit 20 \
-            --json number,title --jq '[.[] | select(.title == env.TITLE)] | .[0].number // empty')"
+          # --limit must comfortably exceed the number of open issues that could carry this label,
+          # or our own older alert falls outside the page, the title filter never sees it, and the
+          # outage path opens a duplicate while recovery leaves the missed one open. This workflow
+          # holds only ONE such issue at a time, so 100 is generous -- but warn loudly rather than
+          # truncate silently if that assumption ever breaks.
+          LISTED="$(gh issue list --label "$LABEL" --state open --limit 100 --json number,title)"
+          if [ "$(jq 'length' <<< "$LISTED")" -ge 100 ]; then
+            echo "::warning::100+ open $LABEL issues — this lookup may be truncated."
+          fi
+          EXISTING="$(jq -r '[.[] | select(.title == $t)] | .[0].number // empty' --arg t "$TITLE" <<< "$LISTED")"
 
           if [ -z "$EXISTING" ]; then
             echo "No open $LABEL issue -- creating one."
@@ -397,8 +409,16 @@ jobs:
 
           # Same exact-title match as the alerting step: this path CLOSES issues, so it must only
           # ever touch the one this workflow opened.
-          EXISTING="$(gh issue list --label "$LABEL" --state open --limit 20 \
-            --json number,title --jq '[.[] | select(.title == env.TITLE)] | .[0].number // empty')"
+          # --limit must comfortably exceed the number of open issues that could carry this label,
+          # or our own older alert falls outside the page, the title filter never sees it, and the
+          # outage path opens a duplicate while recovery leaves the missed one open. This workflow
+          # holds only ONE such issue at a time, so 100 is generous -- but warn loudly rather than
+          # truncate silently if that assumption ever breaks.
+          LISTED="$(gh issue list --label "$LABEL" --state open --limit 100 --json number,title)"
+          if [ "$(jq 'length' <<< "$LISTED")" -ge 100 ]; then
+            echo "::warning::100+ open $LABEL issues — this lookup may be truncated."
+          fi
+          EXISTING="$(jq -r '[.[] | select(.title == $t)] | .[0].number // empty' --arg t "$TITLE" <<< "$LISTED")"
 
           if [ -z "$EXISTING" ]; then
             echo "All endpoints healthy and no open alert issue. Nothing to do."
--- a/.github/workflows/uptime-external.yml
+++ b/.github/workflows/uptime-external.yml
@@ -1,0 +1,357 @@
+# ---------------------------------------------------------------------------------------------
+# WHY THIS WORKFLOW EXISTS -- and why it MUST stay on GitHub-hosted runners.
+#
+# Every other probe we have for Ever Teams (Kubernetes liveness/readiness probes, Uptime Kuma,
+# Prometheus blackbox exporter) runs INSIDE the homelab cluster that also owns our egress uplink,
+# so all of them share a single network fate. On 2026-08-15 a routing fault at the ISP (Orange,
+# Spain) made Cloudflare's 188.114.96.0/22 prefix unreachable *from the homelab only*. ever.team,
+# app.ever.team and api.ever.team all resolve into that exact prefix, so every internal probe went
+# red at once and it looked precisely like "production is down". It was not -- real users on the
+# public internet were served normally the whole time. Hours were burned on a false alarm.
+#
+# This workflow is the independent, OUTSIDE-the-homelab vantage point that would have answered
+# "is it them, or is it us?" in thirty seconds.
+#
+#   >>> DO NOT CHANGE `runs-on: ubuntu-latest` TO A SELF-HOSTED / ARC RUNNER. <<<
+#
+# This repository has a standing "self-hosted runners only" CI policy, and nearly every other
+# workflow here correctly uses the RUNNER_LINUX_X64_* org variables. THIS WORKFLOW IS A
+# DELIBERATE, DOCUMENTED EXCEPTION. Our self-hosted ARC runners live in the same homelab, behind
+# the same uplink, inside the same failure domain as the services being probed. Moving this
+# workflow onto them would not "fix an inconsistency" -- it would silently destroy the only
+# property that makes it worth running, and it would fail us in exactly the incident it was built
+# to catch. GitHub-hosted minutes are free and unlimited on public repositories, so there is no
+# cost argument either. If you are here to "clean this up": please don't.
+# ---------------------------------------------------------------------------------------------
+
+# cspell:ignore namelookup appconnect starttransfer ipify ifconfig blackhole homelab blackbox
+# cspell:ignore mktemp redirs ttfb elif esac endgroup
+
+name: Uptime (external vantage point)
+
+on:
+  schedule:
+    # Every 15 minutes. Cost-safe: this repository is PUBLIC, and GitHub-hosted runner minutes are
+    # free and unlimited on public repositories. The happy path is a handful of curls and finishes
+    # in seconds. If this repo is ever made private, drop this to hourly ('0 * * * *') -- private
+    # repos bill against the 2,000 minutes/month free quota.
+    - cron: '*/15 * * * *'
+  workflow_dispatch:
+
+# Least privilege: read repo metadata, and file/close alert issues. Nothing else.
+permissions:
+  contents: read
+  issues: write
+
+# Never let two probe runs interleave and fight over the same alert issue.
+concurrency:
+  group: uptime-external
+  cancel-in-progress: false
+
+jobs:
+  probe:
+    name: Probe public endpoints
+    # See the banner at the top of this file before touching this line.
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+
+    steps:
+      # No actions/checkout on purpose -- this job needs nothing from the repository tree, and
+      # skipping the checkout keeps each run down to a few seconds.
+      - name: Probe endpoints from outside the homelab
+        id: probe
+        shell: bash
+        run: |
+          set -uo pipefail
+
+          UA='ever-teams-uptime-monitor (+https://github.com/ever-co/ever-teams)'
+
+          # Attempts are spaced so a single transient blip can never raise an alert: an endpoint
+          # is only declared down after ATTEMPTS consecutive failures.
+          ATTEMPTS=3
+          SPACING=20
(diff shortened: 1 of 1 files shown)


evereq:
## Review pass — one blocker, reproduced live

Pushed in `894a3f5` + `df87314`. Every check on the previous head was green and **no bot found the
most important defect**, so this needed a manual pass.

### 🔴 Blocker: the control endpoint silently disarmed every alert

`CONTROL_URL='https://api.github.com/zen'` is unauthenticated, so it is capped at **60 requests/hour
per egress IP**. GitHub-hosted runners share large NAT pools, so it returns
`403 API rate limit exceeded` on ordinary runs — measured directly:

```
HTTP/1.1 403 rate limit exceeded
X-RateLimit-Limit: 60
X-RateLimit-Remaining: 0
```

That sets `control_down=true`, and the alert step is gated on
`failed == 'true' && control_down == 'false'` — so **a real Ever Teams outage would have filed no
issue at all**. A monitor that disarms itself is worse than no monitor. It also burned ~45s of every
run in failed retries, and being GitHub-hosted it could not separate "this runner is broken" from
"the GitHub API is down".

Replaced with `https://checkip.amazonaws.com`: AWS-hosted, so a **Cloudflare**-wide event cannot take
the control down together with all three targets (which are all behind Cloudflare) — `api.ipify.org`
was disqualified for exactly this reason, and was in fact serving a Cloudflare `520` while this was
written. Verified passing on the first attempt.

### Also fixed

- **`https://ever.team/` had no body marker** — only a 1024-byte floor against a real page of
  **59,796 bytes**, 58× headroom. A Cloudflare interstitial, an "Always Online" shell or a
  wrong-environment deploy would all have reported UP. Now asserts `Time Tracking Platform`, proven
  with a known-dirty control (`/bogus-xyz-9931` → 404, marker absent).
- **curl's exit status was discarded** (cubic P2) — a `--max-time` abort still prints
  `http_code=200` with a partial body, so a stalled endpoint could satisfy the marker check.
- **Issue lookups now match the exact title, not the label alone** (cubic/Greptile P1). The premise
  was partly manufactured — the label does not exist yet, so only this workflow's own issues can
  carry it — but the recovery path does not merely comment, it **closes**, and a hand-labelled
  incident issue must never be auto-closed.
- **`EGRESS_IP`** is AWS-first and validated before reaching `$GITHUB_OUTPUT`; a multi-line value
  breaks the file command, fails the step, and every later `if:` carries an implicit `success()` —
  the monitor would go silent rather than loud.
- The **control's row shares the timing table** with the real targets, so a reader saw "check
  passed" directly above a `DOWN` row. Labelled, and a control failure now reports INCONCLUSIVE.
- **"Note that the runner itself is suspect"** was gated on `control_down` alone, so a healthy run
  with a flapping control printed two `::error::` annotations claiming the results were worthless.
- Recovery comment can no longer block the close; `timeout-minutes` 10 → 15.

### Declined

- **cubic P2: gate the final `exit 1` on `control_down == 'false'`.** Dangerous. Before the control
  fix, `control_down` was true on ordinary runs, so this would have reported **green during a real
  outage**. Even with a reliable control it throws away the one always-on notification channel. The
  run stays red; the annotation says INCONCLUSIVE instead.
- **cubic/Copilot: gate the recovery close on `control_down` too.** The control only decides whether
  a *failure* is trustworthy. If all three targets returned 200 with correct markers, the runner
  demonstrably has working reachability — the control adds no doubt. Gating it means resolved
  incidents never auto-close.
- **Copilot/cubic P3: the failing-endpoints list renders as a code block.** Factually wrong — a
  `run: |` block scalar strips the common indent. Disproved twice, including by running the step and
  reading the emitted summary: the bullets sit at column 0. Applying the patch would have broken the
  YAML.
- **FOSSA "License Compliance"** is a branch-level finding on `main`, unrelated to a YAML-only diff.

Everything above was verified by **executing the probe step end-to-end** the way GitHub does
(`bash --noprofile --norc -eo pipefail`), on both the healthy path (4/4 pass) and a known-dirty run
with a bogus marker (`failed=true`).

evereq:
## Second pass — both cubic findings applied (`665e646`)

Both were valid, and **the second was a regression I introduced** in `df87314`. Recording that
plainly rather than quietly fixing it.

**P2, line 264 — marker was marketing copy.** Correct critique. `Time Tracking Platform` lives only
in the `<title>` and `<meta description>`, so a rebrand would take a perfectly healthy site
"DOWN" and open a false alert — the exact failure class this workflow exists to end. Switched to
the semantic `<meta name="author" content="Ever Teams">` tag, verified with a known-dirty
control: present on the live page, absent on `/bogus-xyz-9931` (404).

**P2, line 351 — bounded lookup.** My fault. I introduced `--limit 20` when adding the exact-title
match, which reopened a variant of the very problem the title match was meant to solve: with >20 open
labelled issues our own alert can fall outside the page, so the outage path opens a duplicate and
recovery leaves the missed one open. Raised to 100 with an explicit truncation warning, and the match
now runs through `jq --arg` instead of `env.TITLE`. There are currently **0** open `uptime-alert`
issues so there was no live impact, but the bug was real.

Verified the new lookup offline against decoys — it selects our issue from a list containing an
unrelated hand-labelled issue, and returns empty when only decoys are present — then re-ran the probe
step end-to-end: 4/4 pass.

PR #4433, opened 2026-08-22
ci: route the self-hosted jobs through the shared configure-registry action

evereq:
## What

Replaces the hand-rolled `Configure Registry` step with the shared composite action
`ever-co/ever-gauzy/.github/actions/configure-registry@aa4ee19926fabcf820aa1294385a76aec6bdb548`
in the **20 jobs that run on self-hosted runners**. 17 workflow files touched, `+120 / -222`.

No workflow, job, or step is deleted — every changed site is a like-for-like replacement of one
step by the shared action that serves the same purpose.

## Why

14 of those jobs wrote the **public** route into `.npmrc` unconditionally:

```yaml
if [ -n "${{ secrets.VERDACCIO_TOKEN }}" ]; then
  echo "//packages.ever.co/:_authToken=${{ secrets.VERDACCIO_TOKEN }}" >> .npmrc
  echo "registry=https://packages.ever.co/" >> .npmrc
fi
```

`packages.ever.co` is the Cloudflare-fronted Verdaccio route. It sustains ~280 KB/s and is known to
**truncate large tarballs mid-stream** — measured 2026-08-16 on `@rspack/binding-win32-arm64-msvc@1.7.6`:
9,525,323 bytes from `packages.ever.co` against 14,590,565 bytes byte-exact from both the internal VIP
and npmjs. The truncation is undetectable client-side (Cloudflare re-gzips, so a cut stream is a
complete-looking HTTP 200 whose body fails gunzip), which is how it shows up as a random
`unexpected end of file` or a `sharp` that silently falls back to compiling from source.

The shared action probes the internal Verdaccio VIP (org var `VERDACCIO_REGISTRY`) first, falls back
to public npm when it does not answer, and keeps `packages.ever.co` available behind the
`VERDACCIO_FORCE_PUBLIC` opt-in (no-removal rule — the public path is gated, not deleted).

The other 6 migrated jobs (`extensions.{dev,prod}`, `mobile.apps.{android,ios}`,
`mobile.apps.stage.{android,ios}`) already read `vars.VERDACCIO_REGISTRY`, but only wrote a
`registry=` line. With a committed `yarn.lock` that is close to a no-op: yarn follows the `resolved`
URLs in the lockfile, which still point at `registry.yarnpkg.com`. The shared action rewrites those,
so the internal cache is actually used.

Incidental hardening: the old step spliced `${{ secrets.VERDACCIO_TOKEN }}` straight into the shell
script; the action receives it through `env:`.

## Jobs changed (20)

| Workflow | Job | `runs-on` |
| --- | --- | --- |
| `deploy-vercel-{dev,prod,stage}.yml` | `deploy` | `${{ vars.RUNNER_LINUX_X64_4 \|\| 'ubuntu-latest' }}` |
| `extensions.{dev,prod}.yml` | `build` | `${{ vars.RUNNER_LINUX_X64_4 \|\| 'ubuntu-latest' }}` |
| `mobile.apps.{android,ios}.yml` | `deploy` | `${{ vars.RUNNER_LINUX_X64_4 \|\| 'ubuntu-latest' }}` |
| `mobile.apps.stage.{android,ios}.yml` | `deploy` | `${{ vars.RUNNER_LINUX_X64_4 \|\| 'ubuntu-latest' }}` |
| `mobile.{before-merge,dev,prod,stage}.yml` | `deploy` | `${{ vars.RUNNER_LINUX_X64_4 \|\| 'ubuntu-latest' }}` |
| `web.before-merge.yml` | `deploy` | `${{ vars.RUNNER_LINUX_X64_4 \|\| 'ubuntu-latest' }}` |
| `desktop.apps.yml`, `desktop-server-{api,web}.apps.yml` | `release-linux` | `${{ vars.RUNNER_LINUX_X64_8 \|\| 'ubuntu-latest' }}` |
| `desktop.apps.yml`, `desktop-server-{api,web}.apps.yml` | `release-windows` | `[self-hosted, Windows, X64]` |

`RUNNER_LINUX_X64_4` / `_8` are set at the org (`ever-k8s-linux-x64-4` / `-8`), so these resolve to
the in-cluster ARC pools that can reach the VIP.

## Jobs deliberately NOT changed (9) — they cannot reach the private VIP

Their existing steps are left byte-for-byte as they are, in each of `desktop.apps.yml`,
`desktop-server-api.apps.yml` and `desktop-server-web.apps.yml`:

| Job | `matrix.os` | Why skipped |
| --- | --- | --- |
| `release-mac` | `ghcr.io/cirruslabs/macos-runner:tahoe` | external Cirrus runner, no route to `192.168.1.183` |
| `release-windows-arm64` | `windows-11-arm` | GitHub-hosted |
| `release-linux-arm64` | `${{ vars.RUNNER_LINUX_ARM64_8 \|\| 'ubicloud-standard-8-arm' }}` | external Ubicloud pool; `RUNNER_LINUX_ARM64_8` is unset today so the job is gated off entirely by its `if:` |

`release.sdk.prod.yml` was already migrated to this exact action and pin — untouched.

## Placement / cache safety

The action stays exactly where the old step was: **after** the lockfile-keyed cache restore
(`actions/setup-node` with `cache: yarn`, or `actions/cache@v4` keyed on
`hashFiles('yarn.lock')` in the desktop jobs) and **before** the install.

No cache key needed pinning. Both `actions/setup-node` (`cache-save.ts` reads
`State.CachePrimaryKey`) and `actions/cache@v4` save under the primary key captured at restore
time, not a re-derived `hashFiles()` — so the action's lockfile rewrite cannot bifurcate either
cache. Verified against the upstream sources rather than assumed.

## Verification

- All 43 workflows in `.github/workflows` parse as YAML after the change (0 failures).
- Structural check over the parsed tree: every `Configure Registry` step in a migrated job is a
  `uses:` step on the canonical pin with the three canonical inputs; the 9 skipped jobs still carry
  their original `run:` step (they act as the known-dirty control proving the check discriminates).
- `expect-vip` is derived from the runner rather than hardcoded, matching the idiom already used in
  `release.sdk.prod.yml`: `${{ vars.RUNNER_LINUX_X64_4 != '' }}` / `_8`, and `'true'` for the
  explicit `[self-hosted, Windows, X64]` jobs. It only controls whether an unreachable VIP is
  retried once and warned about; it never fails a job.
- None of these workflows commit or push, so a rewritten `yarn.lock` cannot leak into the repo.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Routes self-hosted CI jobs through the shared `configure-registry` action and re‑pins it to restore a tracked `.yarnrc`, preserving the repo’s `yarn-path` and avoiding slow/truncated installs over the public route.

- Scope and exceptions: 17 self-hosted jobs now use the shared action. The three Vercel deploy jobs keep their inline registry step (they upload the working tree and build on Vercel). `release.sdk.prod` only updates the action pin. External/GitHub-hosted jobs are unchanged.
- Behavior: the action probes `vars.VERDACCIO_REGISTRY`, rewrites lockfile resolved URLs, falls back to public npm, and allows opting into the public route via `vars.VERDACCIO_FORCE_PUBLIC`. `expect-vip` derives from self-hosted Linux runner vars; Windows self-hosted jobs set it to 'true'.
- Mobile EAS uploads: restore `.npmrc`, `.yarnrc`, and `yarn.lock` and remove `*.rewritten` artifacts before `eas build` so VIP URLs/tokens do not ship and Expo installs succeed.
- Setup: ensure `vars.VERDACCIO_REGISTRY` and `secrets.VERDACCIO_TOKEN` exist; use `vars.VERDACCIO_FORCE_PUBLIC` only if needed.
- Lint: extend `.cspell.json` with seven project terms (e.g., EVERDESK, `Jimp`, `vercelignore`) to satisfy `cspell` on changed files; no behavior change.

<sup>Written for commit 451bba911daf1633c13d4e0a21b2d23446129ae6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ever-co/ever-teams/pull/4433?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Improvements**
  - Standardized package registry configuration across development, staging, production, desktop, mobile, web, extension, and SDK workflows.
  - Improved compatibility with remote build environments and dedicated build runners.
  - Added safeguards to restore temporary registry and lockfile changes before mobile submissions.
  - Updated workflow documentation and validation terminology.
  - Existing build, validation, publishing, and submission processes remain unchanged.

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -847,7 +847,14 @@
 		"typeahead",
 		"antfu",
 		"conventionalcommits",
-	],
+		"DLOPEN",
+		"ENOENT",
+		"EVERDESK",
+		"EVERGR",
+		"Jimp",
+		"prebuilts",
+		"vercelignore"
+],
 	"useGitignore": true,
 	"ignorePaths": [
 		"**/locales/**",
--- a/.github/workflows/deploy-vercel-dev.yml
+++ b/.github/workflows/deploy-vercel-dev.yml
@@ -30,12 +30,12 @@ jobs:
           cache: "yarn"
 
       - name: Configure Registry
-        shell: bash
-        run: |
-          if [ -n "${{ secrets.VERDACCIO_TOKEN }}" ]; then
-            echo "//packages.ever.co/:_authToken=${{ secrets.VERDACCIO_TOKEN }}" >> .npmrc
-            echo "registry=https://packages.ever.co/" >> .npmrc
-          fi
+        uses: ever-co/ever-gauzy/.github/actions/configure-registry@aa4ee19926fabcf820aa1294385a76aec6bdb548
+        with:
+          verdaccio-registry: ${{ vars.VERDACCIO_REGISTRY }}
+          verdaccio-token: ${{ secrets.VERDACCIO_TOKEN }}
+          force-public: ${{ vars.VERDACCIO_FORCE_PUBLIC }}
+          expect-vip: ${{ vars.RUNNER_LINUX_X64_4 != '' }}
 
       - name: Install Packages
         run: |
--- a/.github/workflows/deploy-vercel-dev.yml
+++ b/.github/workflows/deploy-vercel-dev.yml
@@ -29,13 +29,33 @@ jobs:
           node-version: "24.14.0"
           cache: "yarn"
 
+      # 🛑 DO NOT migrate this step to the shared ever-gauzy configure-registry action.
+      # It was migrated once and reverted here; this comment is why.
+      #
+      # This job does NOT ship the artifact it builds on the runner. The deploy step below
+      # (BetaHuhn/deploy-to-vercel-action@v1) passes no PREBUILT input and uses
+      # WORKING_DIRECTORY: ./, so Vercel UPLOADS this working tree and runs install + build
+      # REMOTELY on its own builders. There is no .vercelignore in this repo, so every tracked
+      # file goes up exactly as it sits on disk at that moment.
+      #
+      # The shared action mutates three TRACKED files: it appends `registry=<REGISTRY>` to
+      # .npmrc, appends `registry "<REGISTRY>"` to .yarnrc, and sed-rewrites every resolved
+      # URL in yarn.lock to that registry. On these runners <REGISTRY> is the internal
+      # Verdaccio VIP (192.168.1.183:4873), which is LAN-only. Vercel's builders cannot route
+      # to it, so the remote `yarn install` fails on every tarball and the deploy dies.
+      # A post-install restore guard does not help either: it would hand Vercel a config with
+      # no Verdaccio credentials at all.
+      #
+      # The hand-rolled step below is kept ON PURPOSE. It touches only .npmrc, and writes only
+      # the PUBLICLY reachable https://packages.ever.co/ route plus its auth token - so the
+      # config that ships to Vercel still resolves from outside our network.
       - name: Configure Registry
-        uses: ever-co/ever-gauzy/.github/actions/configure-registry@aa4ee19926fabcf820aa1294385a76aec6bdb548
-        with:
-          verdaccio-registry: ${{ vars.VERDACCIO_REGISTRY }}
-          verdaccio-token: ${{ secrets.VERDACCIO_TOKEN }}
-          force-public: ${{ vars.VERDACCIO_FORCE_PUBLIC }}
-          expect-vip: ${{ vars.RUNNER_LINUX_X64_4 != '' }}
+        shell: bash
+        run: |
+          if [ -n "${{ secrets.VERDACCIO_TOKEN }}" ]; then
+            echo "//packages.ever.co/:_authToken=${{ secrets.VERDACCIO_TOKEN }}" >> .npmrc
+            echo "registry=https://packages.ever.co/" >> .npmrc
+          fi
 
       - name: Install Packages
         run: |
--- a/.github/workflows/deploy-vercel-prod.yml
+++ b/.github/workflows/deploy-vercel-prod.yml
@@ -30,12 +30,12 @@ jobs:
           cache: "yarn"
 
       - name: Configure Registry
-        shell: bash
-        run: |
-          if [ -n "${{ secrets.VERDACCIO_TOKEN }}" ]; then
-            echo "//packages.ever.co/:_authToken=${{ secrets.VERDACCIO_TOKEN }}" >> .npmrc
-            echo "registry=https://packages.ever.co/" >> .npmrc
-          fi
+        uses: ever-co/ever-gauzy/.github/actions/configure-registry@aa4ee19926fabcf820aa1294385a76aec6bdb548
+        with:
+          verdaccio-registry: ${{ vars.VERDACCIO_REGISTRY }}
+          verdaccio-token: ${{ secrets.VERDACCIO_TOKEN }}
+          force-public: ${{ vars.VERDACCIO_FORCE_PUBLIC }}
+          expect-vip: ${{ vars.RUNNER_LINUX_X64_4 != '' }}
 
       - name: Install Packages
         run: |
--- a/.github/workflows/deploy-vercel-prod.yml
+++ b/.github/workflows/deploy-vercel-prod.yml
@@ -29,13 +29,33 @@ jobs:
           node-version: "24.14.0"
           cache: "yarn"
 
+      # 🛑 DO NOT migrate this step to the shared ever-gauzy configure-registry action.
+      # It was migrated once and reverted here; this comment is why.
+      #
+      # This job does NOT ship the artifact it builds on the runner. The deploy step below
+      # (BetaHuhn/deploy-to-vercel-action@v1) passes no PREBUILT input and uses
+      # WORKING_DIRECTORY: ./, so Vercel UPLOADS this working tree and runs install + build
+      # REMOTELY on its own builders. There is no .vercelignore in this repo, so every tracked
+      # file goes up exactly as it sits on disk at that moment.
+      #
+      # The shared action mutates three TRACKED files: it appends `registry=<REGISTRY>` to
+      # .npmrc, appends `registry "<REGISTRY>"` to .yarnrc, and sed-rewrites every resolved
+      # URL in yarn.lock to that registry. On these runners <REGISTRY> is the internal
+      # Verdaccio VIP (192.168.1.183:4873), which is LAN-only. Vercel's builders cannot route
+      # to it, so the remote `yarn install` fails on every tarball and the deploy dies.
+      # A post-install restore guard does not help either: it would hand Vercel a config with
+      # no Verdaccio credentials at all.
+      #
+      # The hand-rolled step below is kept ON PURPOSE. It touches only .npmrc, and writes only
+      # the PUBLICLY reachable https://packages.ever.co/ route plus its auth token - so the
+      # config that ships to Vercel still resolves from outside our network.
       - name: Configure Registry
-        uses: ever-co/ever-gauzy/.github/actions/configure-registry@aa4ee19926fabcf820aa1294385a76aec6bdb548
-        with:
-          verdaccio-registry: ${{ vars.VERDACCIO_REGISTRY }}
-          verdaccio-token: ${{ secrets.VERDACCIO_TOKEN }}
-          force-public: ${{ vars.VERDACCIO_FORCE_PUBLIC }}
-          expect-vip: ${{ vars.RUNNER_LINUX_X64_4 != '' }}
+        shell: bash
+        run: |
+          if [ -n "${{ secrets.VERDACCIO_TOKEN }}" ]; then
+            echo "//packages.ever.co/:_authToken=${{ secrets.VERDACCIO_TOKEN }}" >> .npmrc
+            echo "registry=https://packages.ever.co/" >> .npmrc
+          fi
 
       - name: Install Packages
         run: |
--- a/.github/workflows/deploy-vercel-stage.yml
+++ b/.github/workflows/deploy-vercel-stage.yml
@@ -30,12 +30,12 @@ jobs:
           cache: "yarn"
 
       - name: Configure Registry
-        shell: bash
-        run: |
-          if [ -n "${{ secrets.VERDACCIO_TOKEN }}" ]; then
-            echo "//packages.ever.co/:_authToken=${{ secrets.VERDACCIO_TOKEN }}" >> .npmrc
-            echo "registry=https://packages.ever.co/" >> .npmrc
-          fi
+        uses: ever-co/ever-gauzy/.github/actions/configure-registry@aa4ee19926fabcf820aa1294385a76aec6bdb548
+        with:
+          verdaccio-registry: ${{ vars.VERDACCIO_REGISTRY }}
+          verdaccio-token: ${{ secrets.VERDACCIO_TOKEN }}
+          force-public: ${{ vars.VERDACCIO_FORCE_PUBLIC }}
+          expect-vip: ${{ vars.RUNNER_LINUX_X64_4 != '' }}
 
       - name: Install Packages
         run: |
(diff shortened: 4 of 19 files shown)


evereq:
## Adversarial review found three defects; two are now fixed in `d44508ca1`, one is deferred upstream

🛑 **Do not merge this until [ever-co/ever-gauzy#10036](https://github.com/ever-co/ever-gauzy/pull/10036) lands and the pin here is updated.** Reason under Defect 3.

The shared action mutates **three** tracked files, not one: it appends to `.npmrc`, appends `registry "<VIP>"` to `.yarnrc`, and sed-rewrites *every* resolved URL in `yarn.lock` to `192.168.1.183:4873`. The hand-rolled step it replaces touched only `.npmrc`, so every guard written against the old step is now too narrow.

### Defect 1 — EAS upload archive contaminated · FIXED

`eas-cli` ships modified tracked files in the build archive, and `.easignore` lists neither `yarn.lock` nor `.yarnrc` (its line 192 explicitly says those must never be excluded). Expo's cloud builders would have received a lockfile pointing entirely at a LAN IP they cannot route to.

The four `mobile.apps.*` guards now use the loop already proven in `release.sdk.prod.yml`, covering `yarn.lock .npmrc .yarnrc` with `if: always()`.

This was **functionally tested, not eyeballed** — the tree was contaminated exactly as the action does it (**6749** `192.168.1.183` URLs in `yarn.lock`, 3 modified files, `.yarnrc` losing its `yarn-path` pin), then the guard was run verbatim: 0 modified files, 0 VIP references, pin restored.

### Defect 2 — Vercel remote builds · FIXED by reverting to base

`deploy-vercel-{dev,prod,stage}.yml` are now byte-identical to `develop` apart from an explanatory comment (verified by semantic YAML deep-compare, with controls proving the comparator sees real diffs and ignores comments).

A guard would have been **worse** than the revert here: `BetaHuhn/deploy-to-vercel-action@v1` runs with no `PREBUILT` and `WORKING_DIRECTORY: ./`, and no `.vercelignore` exists — so Vercel builds *remotely* from the uploaded tree. The base step deliberately writes the token plus the publicly reachable `packages.ever.co` route so the remote builder can resolve. Stripping that would have left it with no credentials at all.

### Defect 3 — `.yarnrc` destroyed · deferred to the upstream fix

The action runs `rm -f .yarnrc` unconditionally on the stated assumption that it is untracked. **In this repo it is tracked** and carries `yarn-path ".yarn/releases/yarn-1.22.22.cjs"`, so all 20 migrated jobs would install with whatever `yarn` the runner resolves — and `--frozen-lockfile` is yarn-1-only.

Fixing that here would be a workaround; it is fixed at the root in #10036, which moves `.yarnrc` into the action's existing per-path reset loop. **Once #10036 merges, re-pin all 17 sites here to the new sha, then merge.**

### Scope audit — the other 13 jobs need no guard

None export the working tree: the desktop jobs publish built `dist/` installers to GH Releases, the extensions jobs end at `plasmo build --zip`, and the `mobile.*`/`web.before-merge` jobs end at a build. `upload-artifact` appears nowhere in the repo, and `git push` only in `release.sdk.prod.yml` (which already has the correct guard). The verifier separately confirmed the action also rewrites `package-lock.json` — moot here, since this repo has none.

The 9 external-runner jobs (`ubicloud-standard-8-arm`, `cirruslabs/macos-runner`, `windows-11-arm`) remain untouched and action-free, as they must — pointing a runner that can't reach `192.168.1.183` at the VIP hangs the job.

### Note on the red check

`License Compliance` (FOSSA) is **pre-existing on the base branch** — it reports as an `error` commit status on the `develop` HEAD `99f86e04` itself, not something this PR introduced. SonarCloud is likewise red on develop.

evereq:
## Blocker: `deploy` (Mobile Build Before Merge) fails with **ENOSPC**, not a code defect

Run [32596944116](https://github.com/ever-co/ever-teams/actions/runs/32596944116/job/97089330744), step **Install Packages**:

```
error http://192.168.1.183:4873/lightningcss-darwin-x64/-/lightningcss-darwin-x64-1.21.8.tgz:
ENOSPC: no space left on device, write
```

Note the package: **`darwin`**-x64, on a **Linux** runner. This is the documented ever-teams pattern — yarn v1 resolves and downloads *every platform binary* of native `optionalDependencies`, so the yarn cache balloons well past what a typical repo needs. It is exactly why the `-8` pools were bumped 12Gi → 24Gi in k8s-gitops `4dac025`.

**This job runs on `${{ vars.RUNNER_LINUX_X64_4 }}` → `ever-k8s-linux-x64-4`, whose `runner-cache` tmpfs is still 12Gi.** Only the `-8` tier was raised. The run also logged `yarn cache is not found`, i.e. a cold cache — which is when the full download set is fetched and the ceiling is hit.

### What this is *not*

- **Not caused by this PR's diff.** The PR does not change `runs-on` for this job, and it does not change what gets installed. `deploy` is green on `develop`, so the condition is marginal rather than deterministic; a cold-cache run tips it over.
- **Not caused by the tmpfs right-sizing** (k8s-gitops `6a9da19`). That commit touched only `docker-storage` on this pool (24Gi → 16Gi), which is dind's `/var/lib/docker`. A yarn tarball write goes to `runner-cache`, which was left untouched.
- **Not `License Compliance`** — that one is FOSSA and is red on the `develop` HEAD itself, so it is pre-existing and unrelated.

### Why it is worth surfacing rather than just retrying

The package cache is a **memory-backed emptyDir**, so it has a hard ceiling. On disk-backed storage this install just consumes more disk; on tmpfs it fails outright. That ceiling has now forced three bumps (8Gi → 12Gi → 24Gi on `-8`) and is failing again at 12Gi on `-4`.

Measured this session: moving the package cache to RAM produced **no measurable improvement** — bound **−15% to +5%**, with the two high-n probes that actually write into the moved cache showing **+3%/+5%**. The mechanism is that `runner-cache` is an `emptyDir` on an **ephemeral** runner, so the cache starts *empty every job* — it is per-job scratch, not a cache, and those files were previously served from the node page cache for free.

So the options, in preference order, are a matter for the repo/infra owner rather than something to paper over here:

1. **Reconsider the cache-on-tmpfs decision** — it is the root cause, and it has no measured upside.
2. Raise `runner-cache` on the `-4` tier to 24Gi, matching `-8`. Unblocks this, but pins another 12Gi of RAM per runner for a benefit measured at zero.
3. Move this job to the `-8` tier. Narrowest fix, but leaves every other `-4` yarn job on the same cliff.

I have **not** applied any of these — I hold no claim on `arc/values` (released in homelab `c2c7974`), and a concurrent session is actively canarying that area.

**This PR remains blocked on [ever-co/ever-gauzy#10036](https://github.com/ever-co/ever-gauzy/pull/10036) regardless**, after which the pin here must be updated before merge.

evereq:
### Update: the ENOSPC is **systematic**, not a one-off — 2 of 2 cold-cache installs failed

Both `deploy` reds are the same failure, in two independent workflows, on the same pool:

| workflow | runner | failing step | tarball being written |
|---|---|---|---|
| `web.before-merge.yml` | `ever-k8s-linux-x64-4-kng22-…-j2cxg` | Install Packages | `@rolldown/binding-**win32**-x64-msvc` |
| `mobile.before-merge.yml` | `ever-k8s-linux-x64-4-kng22-…-xxhpq` | Install Packages | `lightningcss-**darwin**-x64` |

A **Windows** binary and a **macOS** binary, both on **Linux** runners. That is the signature of yarn v1 resolving every platform variant of native `optionalDependencies`, and it is why the cache blows past 12Gi.

### On whether this PR causes it

It does not cause the *mechanism*, but it is worth being precise about why it surfaced here:

- Both jobs already ran on `${{ vars.RUNNER_LINUX_X64_4 }}` before this PR — `runs-on` is unchanged.
- The same tarball set is downloaded either way; only the host in the URL differs.
- The runs logged `yarn cache is not found` — a genuinely **cold** cache, which is normal for a new PR branch and is exactly when the full platform-variant set is fetched.

I checked the one way this PR *could* have made cold caches more likely — the action rewrites `yarn.lock`, and a cache key derived from the lockfile hash would then bifurcate. It does not apply here: `actions/setup-node`'s post step saves under the key captured at **restore** time (`core.getState(State.CachePrimaryKey)`), it does not re-run `hashFiles()`, and the restore runs *before* Configure Registry. So the cache namespace is intact.

**Conclusion: any PR touching this repo on the `-4` tier will hit this on a cold cache.** It is a pool-sizing issue, not a defect in this diff.

### Why "just raise it to 24Gi" is the wrong reflex

That ceiling exists only because the package cache is a **memory-backed `emptyDir`**. Raising it pins another 12Gi of RAM per runner — and this session measured the benefit of putting that cache in RAM at **no measurable effect** (bound −15% to +5%; the two high-n probes writing into the moved cache came in at **+3%/+5%**). The reason is structural: `runner-cache` is an `emptyDir` on an **ephemeral** runner, so it starts **empty every job** — it is per-job scratch, not a cache, and those files were previously served free from the node page cache.

So the ledger is now **three ENOSPC incidents and three bumps (8→12→24Gi), against a benefit measured at zero.**

Still not applying any fix here — I hold no claim on `arc/values`, and this is an infra decision for the owner. Recorded so it is not rediscovered a fourth time.

evereq:
### Verdict: the ENOSPC **reproduced** — it is confirmed systematic, not a flake

The `chore(cspell)` commit re-triggered every check, which gave a clean reproduction test. Both `deploy` jobs failed again with the identical error:

| run | workflow | result |
|---|---|---|
| 1st | `web.before-merge.yml` | `ENOSPC: no space left on device` |
| 1st | `mobile.before-merge.yml` | `ENOSPC: no space left on device` |
| 2nd (after cspell commit) | `web.before-merge.yml` | `ENOSPC: no space left on device` |
| 2nd (after cspell commit) | `mobile.before-merge.yml` | `ENOSPC: no space left on device` |

**4 failures, 2 workflows, 2 independent runs.** Deterministic on a cold cache for this repo on the `-4` tier. I have deliberately not dispatched a further retry — a fifth reproduction would only consume runner capacity that is already scarce.

### Where that leaves this PR

- ✅ **Cspell fixed** — `336d8b80`. Six of the seven flagged words were **pre-existing on develop** (`EVERDESK` appears 9×, plus `Jimp`, `prebuilts`, `ENOENT`, `DLOPEN`, `EVERGR`); they had simply never been linted because cspell-action only checks *changed* files. Only `vercelignore` came from this PR. Dictionary 843 → 850 words.
- 🔴 **`deploy` ×2 — ENOSPC.** Infrastructure, not this diff: `runs-on` is unchanged, the same tarball set is fetched either way, and the cache-key path was checked and is not bifurcated.
- 🔴 **`License Compliance`** — FOSSA, red on the `develop` HEAD itself. Pre-existing.

### The decision this surfaces

The package cache is a **memory-backed `emptyDir`**, so it has a hard ceiling. Three bumps have already been spent chasing it (8 → 12 → 24Gi on `-8`), and `-4` is now failing at 12Gi. Measured this session, the benefit of putting that cache in RAM is **no measurable effect** — bound −15% to +5%, with the two high-n probes that write into the moved cache at **+3%/+5%**. `runner-cache` is an `emptyDir` on an **ephemeral** runner, so it starts empty every job: it is per-job scratch, not a cache.

So the honest options, none of which I am taking unilaterally:
1. **Move the package cache off tmpfs** — root cause, and the measurement says nothing is lost.
2. Raise `-4` to 24Gi — unblocks this, pins another 12Gi of RAM per runner for a benefit measured at zero.
3. Move these two jobs to `-8` — narrowest, leaves every other `-4` yarn job on the same cliff.

**This PR is still gated on [ever-co/ever-gauzy#10036](https://github.com/ever-co/ever-gauzy/pull/10036)** (the `.yarnrc` restore fix) and must be re-pinned to the new action sha before merge, so there is no urgency in picking an option tonight.

